PR #4, opened 2016-06-13
Genericize JavaScript for optional sections

gtvj:
Separates the JavaScript functionality for:
- Showing/hiding optional sections
- Removing optional sections

Into two jQuery plugins that can be used in isolation or in tandem
depending on specific needs.


dan-palmer:
@gtvj is this awaiting tests?

@dan-palmer it is. Currently having some problems with Jasmine fixtures (think it might be the Gem we're using). I'm currently working on fixing this.

PR #7, opened 2016-06-16
Healthcare section

cesidio:


dan-palmer:
@cesidio ping me when this is ready for review

app/form_builders/mps_form_builder.rb
```diff
@@ -0,0 +1,14 @@
+class MpsFormBuilder < GovukElementsFormBuilder::FormBuilder
+  def radio_with_toggle(attribute)
+    style = 'optional-section-wrapper panel panel-border-narrow'
+    content_tag(:div, class: 'js_multiples') do
+      safe_join([
+        content_tag(:div, class: 'form-group controls-optional-section') do
+          radio_button_fieldset attribute,
+            choices: object.toggle_choices, inline: true
+        end,
+        content_tag(:div, class: style) { text_area "#{attribute}_details" }
```
dan-palmer: I think this should be generic and just yield a block. Maybe have a specific textarea form method as well.

PR #17, opened 2016-06-24
Fit establishment and must return in one line

cesidio:


app/cells/move_information/show.slim
```diff
@@ -21,20 +21,22 @@
...
+          .form-group
+            - f.fields_for :destinations do |df|
+              .form-group.optional-section.panel.panel-border-narrow
+                = df.hidden_field :id
+                .form-group
+                  .establishment-field
+                    = df.text_field :establishment
+                  .must-return-field
+                    = df.label_with_radio(:must_return, 'Must return', 'must_return')
+                    = df.label_with_radio(:must_return, 'Must NOT return', 'must_not_return')
```
whoojemaflip, line 33: should our UI strings all be I18N?

cesidio, line 33: We have a ticket in this sprint for introducing all the copy in the translation file

PR #20, opened 2016-06-27
Multiples focus

cesidio:


dan-palmer:
The jasmine specs should be running on circle too.

PR #26, opened 2016-06-30
adds ignore rule for spec/examples.txt

whoojemaflip:


dan-palmer:
👍 just needs autosquashing

PR #27, opened 2016-06-30
configure local database host

whoojemaflip:
As a postgres.app user, I need this config to allow Rails to do what it does. Hopefully it won't break anything for yous.


dan-palmer:
👍 all good

PR #34, opened 2016-07-01
Introduce risks controller with risks to self form

cesidio:


app/controllers/concerns/wizardable_controller.rb
```diff
@@ -0,0 +1,33 @@
+module WizardableController
```
whoojemaflip: since this isn't a controller, should it be called x_controller?

app/views/healthcare/_allergies.html.slim
```diff
@@ -1,4 +1 @@
-= form_for form, url: form_path, method: :put do |f|
-  = f.radio_toggle_with_textarea :allergies
-
-  = render partial: 'buttons', locals: { f: f }
+= f.radio_toggle_with_textarea :allergies
```
whoojemaflip, line 1: why does this PR affect the healthcare partials?

cesidio, line 1: Should have been done in a different PR, but I removed an helper method from the controller and made the healthcare views more consistent with the ones we have in risks

PR #48, opened 2016-07-05
Offences controller

whoojemaflip:


app/views/offences/show.html.slim
```diff
@@ -0,0 +1,11 @@
...
+  header
+    h1 Offences
+
+  = GovukElementsErrorsHelper.error_summary form,
+    'There were problems saving this form', ''
+
+  = form_for form, url: offences_path(@escort), method: :put do |f|
+    = f.text_field :release_date
+
+    = f.submit 'Save and continue', class: 'button'
```
cesidio: Only this missing new line and then 👍

PR #58, opened 2016-07-07
Complete offences form

whoojemaflip:
It's not 100% by any means - the invalid form states are fugly, and there's no feature spec yet.

Posted so you can see the progress.


cesidio:
👍

PR #59, opened 2016-07-08
Implement flow

cesidio:
This PR will add the required flow when we search for a detainee with:

no move
past move
future move

showing the proper results from the search in the home page, with the action to add a new move or view profile.


app/models/escort.rb
```diff
@@ -29,4 +29,16 @@ def offences
   def risks
     super || build_risks
   end
+
+  def with_future_move?
+    true if move.date && move.date > Date.current
```
whoojemaflip: There's no need for the `true if` - this method can just return the boolean outcome of the logical &&

app/models/forms/base.rb
```diff
@@ -120,10 +120,6 @@ def prepopulated_collection(field_name)
       end
     end
 
-    def invalid?
```
whoojemaflip, line 123: This change doesn't seem to be connected to the rest of the PR

spec/factories/detainee_factory.rb
```diff
@@ -0,0 +1,13 @@
+FactoryGirl.define do
+  factory :detainee do
+    forenames 'Donald'
```
whoojemaflip, line 3: Prefer fictional names in test data!

PR #60, opened 2016-07-11
A test for prepopulated collections (multiples)

whoojemaflip:
It's maybe not 100% complete, but seems like a reasonable start to the issue of testing the complexity of Multiples in their current incarnation.


cesidio:
👍  Nice work Tom

PR #61, opened 2016-07-11
Use db/seeds.rb for preseeding database

whoojemaflip:
There is already a rake task `rake db:seed` for preseeding the DB with data e.g. a test user.

It seems unobvious to have a separate task for this purpose.


cesidio:
https://github.com/thoughtbot/guides/tree/master/best-practices

Use db/seeds.rb for data that is required in all environments.
Use dev:prime rake task for development environment seed data.

We followed the tip of not using the standard db/seeds.rb file as this data is used only in development, that was the rationale behind it.

whoojemaflip:
Okay, I can see the justification. Seemed weird that's all.

dan-palmer:
Using seeds seems a little more like expected behaviour in most circumstances. You also get the user added to the db for free if you setup with bin/setup or rake db:setup.

dev:prime is just something else to remember.

We don't really need to seed the db with anything on deploy in prod so the argument for separating prod vs dev things is a kinda void - you can just test for dev in the seeds script. 

Also not a fan of the word prime, it's masquerading as a synonym to setup - which is already the term for this sort of behaviour. 

+1 for simplifying/normalising behaviour.

PR #62, opened 2016-07-11
Change button to link in homepage search result and order of validation fields

cesidio:


app/models/forms/risks/communication.rb
```diff
@@ -1,14 +1,14 @@
...
-      optional_details_field :can_read_and_write
-
       optional_field :interpreter_required
       property :language, type: StrictString
       validates :language,
         presence: true,
         if: -> { interpreter_required == TOGGLE_YES }
+
+      optional_details_field :hearing_speach_sight
+      optional_details_field :can_read_and_write
```
whoojemaflip, line 11: why have these properties moved?

cesidio, line 11: Because the order affects the error messages on top of the page. Matt wants the list of errors you read on top of the page reflect the order in which these fields are presented on the page.

whoojemaflip, line 11: fair enough.

whoojemaflip:
:+1:

PR #64, opened 2016-07-11
Add detainee details on profile page

cesidio:


app/presenters/detainee_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+  def short_gender
+    if gender == 'male'
+      'M'
+    elsif gender == 'female'
+      'F'
+    end
+  end
+
+  def age
+    # TODO: Add validation on date_of_birth and remove presence check
```
whoojemaflip: what's with the TODO?

cesidio: All the fields in the detainee must be provided, whether if they come from NOMIS or by the prison staff. This is just a remainder that we have to implement that logic sooner or later. They will ask for it.

whoojemaflip: Surely that will live in the Form object, not the presenter?

whoojemaflip:
:+1:

PR #80, opened 2016-07-15
Workflow

whoojemaflip:


app/controllers/healthcare_controller.rb
```diff
@@ -22,6 +23,12 @@ def update
 
   private
 
+  def update_document_workflow
```
cesidio, line 28: What about moving this method to application_controller?

whoojemaflip, line 28: We're going to add a few more of these calls to our controllers soon. Are you happy to wait and see how that pans out before refactoring?

cesidio, line 28: sure.

happy with the PR for the rest.

PR #81, opened 2016-07-18
be more RESTful

whoojemaflip:
Rendering content on a POST request is gross, and having routes that don't resolve is bad too. Merging this will tidy up some loose ends  in that regard.


cesidio:
👍

PR #82, opened 2016-07-18
Escort.create_with_children

whoojemaflip:


app/controllers/escorts_controller.rb
```diff
@@ -3,10 +3,7 @@ def create
     form = Forms::Search.new
 
     if form.validate(prison_number: params[:prison_number])
-      escort = Escort.new
-      escort.build_detainee(prison_number: form.prison_number)
-      escort.save!
-
+      escort = Escort.create_with_children(prison_number: form.prison_number)
```
dan-palmer, line 6: 🎉

app/models/escort.rb
```diff
@@ -9,20 +9,18 @@ def self.find_detainee_by_prison_number(number)
...
+    end
   end
 
-  def offences
-    super || build_offences
-  end
-
-  def risks
-    super || build_risks
+  def move
```
dan-palmer, line 22: @whoojemaflip How come this one is lazy?

whoojemaflip, line 22: It's got different semantics. Not quite sure what they are yet, but it doesn't fall into the exact same category as Risks or Healthcare.

PR #83, opened 2016-07-18
Healthcare and risk section summary pages

cesidio:


app/assets/stylesheets/_helper_classes.scss
```diff
@@ -54,3 +65,15 @@ label[for$=_unknown] {
 .text-error {
   color: red;
 }
+
+.pull-right-content {
+  float: right;
+}
+
+.complete-label {
```
whoojemaflip: do these need to be in the top-level namespace for our CSS? Maybe it's time to have the BEM conversation

spec/presenters/summary_presenter_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
```
whoojemaflip: does this unit test really need to include Rails?

spec/presenters/summary_presenter_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+RSpec.describe SummaryPresenter, type: :model do
```
whoojemaflip: is this actually a `:model` test or a unit test of something else (a presenter)?

app/assets/stylesheets/_helper_classes.scss
cesidio: Sure happy to have a discussion about it. It may apply to other places as well. Shell we have it now or when in a less busy period?

whoojemaflip:
:+1:

PR #86, opened 2016-07-19
Provide missing csra icon

dan-palmer:
Turns this:

screen shot 2016-07-19 at 15 45 44

into:

screen shot 2016-07-19 at 15 45 25


whoojemaflip:
:+1:

PR #89, opened 2016-07-20
Show must return and must not return on profile

cesidio:


spec/presenters/move_presenter_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe MovePresenter, type: :model do
+RSpec.describe MovePresenter, type: :presenter do
```
dan-palmer, line 3: 👍

PR #92, opened 2016-07-20
Spec optional fields

dan-palmer:
- [x] Optional field spec helper
- [x] Apply to all forms that use the optional_field helper(ignoring optional_details_field for now - will handle with its own matcher)


whoojemaflip:
🔢

PR #94, opened 2016-07-20
DocumentWorkflow interrogators

whoojemaflip:
helper methods to interrogate workflow_state

### DIFF
--- a/spec/services/document_workflow_spec.rb
+++ b/spec/services/document_workflow_spec.rb
@@ -49,7 +49,7 @@ RSpec.describe DocumentWorkflow do
     let(:result) { subject.update_status(new_status) }
 
     context "with an invalid workflow state" do
-      let(:new_status) { 'monkies' }
+      let(:new_status) { :monkies }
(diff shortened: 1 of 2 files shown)


cesidio:
👍

spec/services/document_workflow_spec.rb
```diff
@@ -1,7 +1,49 @@
...
+  let(:initial_status) { 'not_started' }
+
+  describe "state_interrogator" do
+    shared_examples "state_interrogator" do |state|
+      let(:result) { subject.public_send("is_#{state}?") }
+
+      context "model is in #{state} state" do
+        let(:initial_status) { state }
+
+        it "is true" do
```
cesidio: what about

its(:result) { is_expected.to be true }

dan-palmer: A similar alternative without relying on the its gem is RSpecs specify

dan-palmer:
:+1:

PR #97, opened 2016-07-21
Change the toggle field type to String

dan-palmer:
We don't need to alter the coercion strategy on empty strings
for the toggle fields because the value can be only one of three
available flags: yes, no & unknown.

This is a precursor to improving the form macros in all specs.


cesidio:
👍

whoojemaflip:
as long as the macro helper asserts the value of the form element, this is great. 🐑  🇮🇹

PR #98, opened 2016-07-21
Guard filter on all document actions

whoojemaflip:
This controls access to all viewing and editing of document summary and edit forms, according to the AccessPolicy.


cesidio:
💯

PR #99, opened 2016-07-21
Remove old specs that are now handled by a matcher

dan-palmer:
Redundant code, the handy `validate_prepopulated_collection` now handles this part of the spec.


cesidio:
👍

whoojemaflip:
good catch

PR #100, opened 2016-07-21
Feature for preliminary Print work

whoojemaflip:


dan-palmer:
👍

app/views/homepage/_moves.html.slim
```diff
@@ -18,7 +18,7 @@
           span.strong-text
             | #{move.detainee.surname}
           span
-            |  #{move.detainee.forenames}
+            | #{move.detainee.forenames}
```
cesidio: In order to preserve the white space here, we can use slim:
https://github.com/slim-template/slim#verbatim-text-with-trailing-white-space-

```
span.strong-text
  ' #{move.detainee.surname}
span
  | #{move.detainee.forenames}
```

PR #102, opened 2016-07-21
Add document workflow to profile and summary pages

cesidio:


app/controllers/healthcare_controller.rb
```diff
@@ -27,11 +27,17 @@ def summary
     render 'summary/healthcare'
   end
 
+  def confirm
+    workflow = DocumentWorkflow.new(healthcare)
+    workflow.update_status(:complete)
```
whoojemaflip: `DocumentWorkflow#update_status` can return `false` if the new document status isn't possible. What kind of error handling do you think there should be here?

app/controllers/offences_controller.rb
```diff
@@ -22,8 +22,7 @@ def update
 
   def update_document_workflow
     workflow = DocumentWorkflow.new(offences)
-    workflow.update_status(:complete) ||
-      workflow.update_status(:incomplete)
+    workflow.update_status(:complete)
```
whoojemaflip: This could maybe use a comment to highlight that _currently_ the Offences document is considered to be complete if the form passes validation.

app/models/concerns/workflowable.rb
```diff
@@ -0,0 +1,17 @@
+module Workflowable
```
whoojemaflip: There's another way of doing this, that doesn't extend the models. `DocumentWorkflow.new(risk).is_incomplete?`

app/views/profiles/_status.html.slim
```diff
@@ -0,0 +1,16 @@
+.float-right
```
whoojemaflip: I would prefer this to be a `Block` in BEM terminology, rather than using a global class to define it's layout.

app/controllers/healthcare_controller.rb
cesidio: The front-end prevents showing the button that reaches this endpoint unless the section is in needs_review, and you can get there only when all questions are answered. 
`- if healthcare.needs_review?
    = render partial: 'summary/confirm_and_save', locals: { path: confirm_healthcare_index_path }
`

app/views/offences/show.html.slim
```diff
@@ -1,6 +1,13 @@
 .offences
-  header
-    h1 Offences
+  - if offences.needs_review?
+    header
+      h1 Review before you save and continue
+    h3
+      ' Status
```
dan-palmer, line 6: cool

app/controllers/healthcare_controller.rb
whoojemaflip: yeah makes sense.

app/controllers/risks_controller.rb
```diff
@@ -27,11 +27,17 @@ def summary
     render 'summary/risk'
   end
 
+  def confirm
+    workflow = DocumentWorkflow.new(risk)
```
whoojemaflip: if we're not checking for the outcome of this, it could be a one-liner:

`DocumentWorkflow.new(risk).update_status(:complete)`

dan-palmer: 👍 applies to `HealthcareController` too.

app/controllers/offences_controller.rb
```diff
@@ -22,8 +22,8 @@ def update
 
   def update_document_workflow
     workflow = DocumentWorkflow.new(offences)
-    workflow.update_status(:complete) ||
-      workflow.update_status(:incomplete)
+    # Differently from other sections, there is not an incomplete status for offences
+    workflow.update_status(:complete)
```
whoojemaflip: and here!

whoojemaflip:
:100:

PR #104, opened 2016-07-21
StrictString matcher

dan-palmer:
This is a replacement matcher for `nilify_empty_strings_for`.

Extracting this into a class allows easier reuse in other
custom matchers.

The name has been changed to better reflect its association
with the StrictString form type.


whoojemaflip:
:ok:

PR #107, opened 2016-07-25
Add timestamp fields to Risks table

dan-palmer:
❗ this will break if just running a plain old `rails db:migrate` - you'll need to setup from scratch.


db/schema.rb
```diff
@@ -105,9 +105,9 @@
     t.uuid     "escort_id"
     t.date     "release_date"
     t.boolean  "not_for_release"
+    t.text     "not_for_release_details"
```
dan-palmer, line 108: 😕

whoojemaflip:
This will presumably break staging then?

PR #111, opened 2016-07-26
Ensure factory girl collections aren't empty

dan-palmer:
When we ask for a collection we should expect at least
one element. This sets up a random number in a given
collection between 1 & 5.


whoojemaflip:
🍏

PR #126, opened 2016-08-04
Remove open acct details

dan-palmer:
Based on feedback from the users open_acct_details is no longer required. 🙌

screen shot 2016-08-04 at 12 44 11


whoojemaflip:
:sheep:

PR #127, opened 2016-08-04
Only show CSRA details when high option is selected

dan-palmer:
❗️If JS is off/unavailable the box will be visible.

screen shot 2016-08-04 at 13 55 26

screen shot 2016-08-04 at 13 56 31

### DIFF
--- a/app/models/forms/risk/risk_from_others.rb
+++ b/app/models/forms/risk/risk_from_others.rb
@@ -14,7 +14,7 @@ module Forms
         property(:csra_details, type: StrictString)
         validates :csra_details,
           presence: true,
-          if: -> { csra == CSRA_HIGH || csra == CSRA_STANDARD }
+          if: -> { csra == CSRA_HIGH }
       end
 
       optional_details_field :rule_45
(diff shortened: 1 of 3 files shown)


app/models/forms/risk/risk_from_others.rb
```diff
@@ -1,22 +1,27 @@
 module Forms
   module Risk
     class RiskFromOthers < Forms::Base
-      CSRA_TOGGLE_CHOICES = %w[ high standard unknown ]
+      CSRA_HIGH = 'high'
+      CSRA_STANDARD = 'standard'
+      CSRA_TOGGLE_CHOICES = [CSRA_HIGH, CSRA_STANDARD, DEFAULT_CHOICE]
+
+      def self.configure_csra_section
```
whoojemaflip: does this need to be a method?

dan-palmer: That commit was just an attempt at grouping(showing they have a relationship) those two things which feel kind of inextricably linked in the form and as the CSRA section is slightly different it can't be configured using the macro. Alternative way of achieving the same thing could be to use something like ```concerning :ConfigureCsraSection do ..` or left alone.

This addition was partly influenced by missing the details of this particular section of the form when I was building the reset attribute stuff the other day.

PR #130, opened 2016-08-05
adds indexes on foreign keys

whoojemaflip:
Doesn't fix the glaring performance problem on the homepage, but nonetheless this needed doing.


dan-palmer:
🚢 🚢 🚀

PR #134, opened 2016-08-05
Remove CSRA 'no' image

dan-palmer:
The allowed flags do not include 'no'.

Currently they can only be: **high**, **standard** & **unknown**.


whoojemaflip:
:art:

PR #135, opened 2016-08-05
Cannot create Move in the past

whoojemaflip:


app/models/forms/moves/information.rb
```diff
@@ -31,7 +31,11 @@ class Information < Forms::Base
 
       def validate_date
         # TODO: extract a common date validator
-        errors.add(:date) unless date.is_a? Date
+        if date.is_a? Date
+          errors[:date] << 'Date must not be in the past.' if date < Date.today
```
dan-palmer, line 35: 🙏

spec/forms/moves/information_spec.rb
```diff
@@ -98,6 +98,20 @@
...
+      context "with a date in the past" do
+        it 'returns false' do
+          params[:date] = '01/01/2015'
+          expect(subject.validate(params)).to be false
+        end
+
+        it 'sets a descriptive error on date' do
+          params[:date] = '01/01/2015'
+          subject.validate(params)
+          expect(subject.errors).to include :date
```
dan-palmer: Might not need this check as the one beneath validates that `:date` is present in order to return the error msg.

PR #115, opened 2016-07-29
Reset attributes nested under toggle fields

dan-palmer:
- [x] implement reset macro
- [x] apply reset to optional details field
- [x] apply to misc risk toggle fields
- [ ] reset csra - this one is slightly different to normal toggle fields
- [ ] reset sex offence details - this one is slightly different to normal toggle fields
- [ ] reset Arson details
- [ ] specs needs sorting out
- [ ] checkboxes with details fields

PR #144, opened 2016-08-12
Implement reset functionality

dan-palmer:
First of a few PR's to introduce resetting attribute values. Breaking down #115.

This just introduces the reset macro & tests the abstract form.

### DIFF
--- a/app/models/forms/attribute_reset_collection.rb
+++ b/app/models/forms/attribute_reset_collection.rb
@@ -7,16 +7,18 @@ module Forms
       @collection = []
     end
 
+    attr_reader :collection
+
     def add(*args)
-      @collection << ResetData.new(*args)
+      collection << ResetData.new(*args)
     end
 
     def any?
-      @collection.any?
+      collection.any?
     end
 
-      @collection.each do |reset_obj|
     def reset_all(form, defaults)
+      collection.each do |reset_obj|
         next if form.public_send(reset_obj.master_attribute) == reset_obj.enabled_value
 
         reset_obj.attributes_to_reset.each do |attribute|
--- a/app/models/forms/attribute_reset_collection.rb
+++ b/app/models/forms/attribute_reset_collection.rb
@@ -1,5 +1,6 @@
 module Forms
   class AttributeResetCollection
+    # ([Symbol::Attribute], Symbol::Attribute, String) => Struct::ResetData
     ResetData = Struct.new(:attributes_to_reset, :master_attribute, :enabled_value)
 
     def initialize
@@ -10,12 +11,16 @@ module Forms
       @collection << ResetData.new(*args)
     end
 
-    def perform(form)
+    def any?
+      @collection.any?
+    end
+
+    def perform(form, defaults)
       @collection.each do |reset_obj|
         next if form.public_send(reset_obj.master_attribute) == reset_obj.enabled_value
 
         reset_obj.attributes_to_reset.each do |attribute|
-          form.public_send("#{attribute}=", nil)
+          form.public_send("#{attribute}=", defaults[attribute])
         end
       end
     end
--- a/app/models/forms/attribute_reset_collection.rb
+++ b/app/models/forms/attribute_reset_collection.rb
@@ -1,0 +1,23 @@
+module Forms
+  class AttributeResetCollection
+    ResetData = Struct.new(:attributes_to_reset, :master_attribute, :enabled_value)
+
+    def initialize
+      @collection = []
+    end
+
+    def add(*args)
+      @collection << ResetData.new(*args)
+    end
+
+    def perform(form)
+      @collection.each do |reset_obj|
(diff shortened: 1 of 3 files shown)


app/models/forms/attribute_reset_collection.rb
```diff
@@ -0,0 +1,23 @@
...
+    def add(*args)
+      @collection << ResetData.new(*args)
+    end
+
+    def perform(form)
+      @collection.each do |reset_obj|
+        next if form.public_send(reset_obj.master_attribute) == reset_obj.enabled_value
+
+        reset_obj.attributes_to_reset.each do |attribute|
+          form.public_send("#{attribute}=", nil)
```
whoojemaflip: should the attribute ever be reset to !nil?

spec/forms/base/resetting_attributes_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+RSpec.describe Forms::Base, "resetting attributes" do
+  let(:form) do
+    Class.new(described_class) do
+      property :foo, type: String, validates: { presence: true }
+      property :bar, type: String
+
+      reset attributes: %i[ bar ], if_falsey: :foo, enabled_value: "yes"
+
+      def self.name; "LOL"; end # form base will keel over and die without it
```
whoojemaflip, line 11: 😆

app/models/forms/attribute_reset_collection.rb
dan-palmer: Good point. For text fields nil seems to be the right choice. For the radio buttons I've made the assumption based on the schema.rb, it contains defaults for the text toggle fields but not anything else that is nested.

whoojemaflip: this is where its a bit weird. The forms give a default value, not the models or schema

dan-palmer: I did consider that, but didn't go anywhere with it when I realised we prescribe choices for some of the fields already(e.g. arson & csra) & this just felt like an extension of that. It kind of feels like theres either a small layer missing or the form objects are a little beyond what you'd expect from a form object.

whoojemaflip: do you think that fields should be given a default value in the schema then?

dan-palmer: We could have it look at the schema with something like this:

``` ruby
form.model.column_defaults[attribute]
```

Applied to the reset collection thing it might look something like this:

``` ruby
class AttributeResetCollection
    ...
    def perform(form)
      @collection.each do |reset_obj|
        next if form.public_send(reset_obj.master_attribute) == reset_obj.enabled_value

        reset_obj.attributes_to_reset.each do |attribute|
          reset_value = form.model.column_defaults[attribute]
          form.public_send("#{attribute}=", reset_value)
        end
      end
    end
end
```

whoojemaflip: how about the default value is passed in to the `ResetData` struct by the form itself?
`ResetData = Struct.new(:attributes_to_reset, :master_attribute, :enabled_value, :default_value)`

```
concerning :ResetAttributes do
  included do
    class << self
      def reset(attributes:, if_falsey:, enabled_value: TOGGLE_YES)
        default_value = model.column_defaults[attribute]
        resettable_attributes.add(attributes, if_falsey, enabled_value, default_value)
      end
    end
  end
end
```

That would mean the `AttributeResetCollection` didn't need to know so much about the form's API

dan-palmer: +1 for DI with the defaults. Did attempt that but the `model` isn't available in the class method 😢.  Instead I moved the passing to the `perform` method. Tell me whats ya fink. No rush, can discuss on Monday - have a good weekend dude.

app/models/forms/base.rb
```diff
@@ -12,6 +12,32 @@ class Base < Reform::Form
...
+
+          def resettable_attributes
+            @_resettable_attributes ||= Forms::AttributeResetCollection.new
+          end
+        end
+
+        def validate(*)
+          super.tap do |valid|
+            if valid && self.class.resettable_attributes.any?
+              self.class.resettable_attributes.perform(self, default_attribute_values)
```
whoojemaflip: `.reset_all`?

spec/forms/base/resetting_attributes_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "rails_helper"
+
+RSpec.describe Forms::Base, "resetting attributes" do
+  let(:form) do
+    Class.new(described_class) do
+      property :foo, type: String, validates: { presence: true }
+      property :bar, type: String
+
+      reset attributes: %i[ bar ], if_falsey: :foo, enabled_value: "yes"
```
whoojemaflip, line 9: Any plans to create an RSpec matcher to test for conformance of specific form instances?

PR #157, opened 2016-08-18
Adds a DB backed session store

dan-palmer:
We don't render on POST/PUT requests, instead we redirect the user.
In order to get access to invalid form objects on redirects when
invalid we need to access the session. A cookie isn't big enough, we
can have more than 4KB of data in some instances which will raise
a cookie overflow error. DB sessions solve this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 ruby '2.3.0'
 
+gem 'activerecord-session_store'
 gem 'rails', '5.0.0'
 gem 'pg', '~> 0.18'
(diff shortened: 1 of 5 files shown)


whoojemaflip:
Theres a slight problem with the active-record session store which is that the sessions table doesn't naturally expire stale data.  What mechanism do you propose to prevent runaway growth on this table?

whoojemaflip:
maybe we shouldn't worry too much about the table growth. There's a rake task to prune stale session data, we just need to run it occasionally.  

🐑 🇮🇹

PR #162, opened 2016-08-22
Provide a custom sex offences summary presenter

dan-palmer:
Sex offences doesn't behave like there questions in the
sense that most are a toggle with a text field. Instead it
has a category and a details field(depending on the type of
category) its output on the summary needs to reflect this.

### DIFF
--- a/app/presenters/summary/sex_offences_presenter.rb
+++ b/app/presenters/summary/sex_offences_presenter.rb
@@ -1,0 +1,21 @@
+module Summary
+  class SexOffencesPresenter < SummaryPresenter
+    def details_for(*)
+      "#{victim_type}. #{details}" if sex_offence_checked?
+    end
+
+    private
+
+    def sex_offence_checked?
+      public_send(:sex_offence) == 'yes'
+    end
+
+    def victim_type
+      public_send(:sex_offence_victim).humanize
+    end
+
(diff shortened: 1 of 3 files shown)


spec/presenters/summary/risk/sex_offences_presenter_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.describe Summary::Risk::SexOffencesPresenter, type: :presenter do
+  let(:model) { instance_double(Risk, sex_offence: 'unknown') }
+  subject { described_class.new(model) }
+
+  # it already implements the answer_for method etc
+  it { is_expected.to be_a SummaryPresenter }
+
+  describe '#details_for' do
+    describe 'accepts an argument in order to appear like a regular summary presenter' do
```
dan-palmer: FML 😭

app/presenters/summary/risk/sex_offences_presenter.rb
```diff
@@ -0,0 +1,23 @@
+module Summary
+  module Risk
```
whoojemaflip: does this need such a deeply nested namespace?

whoojemaflip:
📦 :accept:

PR #165, opened 2016-08-25
Considerations concern

whoojemaflip:
introduces Considerations to the model layer. The intent here is to move information about model semantics out of the form layer, so it is queryable and reusable. The short-term goal of this is to reduce duplication of this info across the codebase. Longer term, it is hoped this will make it easier to reason about the relationships between schema fields.


dan-palmer:
👍

whoojemaflip:
-= DON'T MERGE =- 

Replacement in progress

PR #168, opened 2016-09-01
Behavioural analytics

whoojemaflip:
adds JSON logging for date changes via the dashboard. It is intended that Kibana will be configured to graph this data (in the short term)


dan-palmer:
🌞

PR #170, opened 2016-09-02
Infer template name from the current step

dan-palmer:
It's ok for the template name we are rendering to be coupled
to the controller(this stuff already lives there) just not to the form object.


whoojemaflip:
😭

PR #171, opened 2016-09-08
Partial rewrite of Models and Forms to new Considerations pattern

whoojemaflip:
This was a spike that got out of hand. The implementation is mostly there, with the few remaining bugs mainly in the Features / Page Objects code. 

The intention here is to make it easier to make changes to the document model in the future. There is significant de-duplication of concerns, and a concentration of knowledge in the models.

Owing to developer turnover, this PR is not yet suitable for merging as-is - further work required to make it production ready.


lostie:
Not suitable for merge, keeping the branch only as reference.

PR #179, opened 2016-11-02
Fix minor deprecation warning related with Rubocop

lostie:
Whenever running a `rake` command a deprecation warning is currently being displayed:

```bash
> bundle exec rake -vT
[DEPRECATION] `last_comment` is deprecated.  Please use `last_description` instead.
rake about                              # List versions of all Rails frameworks and the environment
```

The offending gem causing this issue seems to be `rubocop` as mentioned in the [following post](https://codedump.io/share/j6nZB0MLIIlA/1/how-to-detect-what-causes-a-deprecation-warning-in-rake)

This updates the gem (which consequently) solves the deprecation warning. 

Due to the nature of the update, I **had to fix some of the new Rubocop offences**. I've intentionally left some of them disabled (for now) as they need a little bit more look into.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,3 +61,9 @@ Style/PredicateName:
 
 Style/MultilineOperationIndentation:
   EnforcedStyle: indented
+
+Style/FrozenStringLiteralComment:
+  Enabled: false
+
+Style/IndentArray:
+  Enabled: false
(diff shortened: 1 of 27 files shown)


lostie:
Seems like the update of the gem causes Rubocop to raise some offences which were not being before.

Looking into it 👓

lostie:
Fixed most of the Rubocop offences. Refer to the (updated) description for further details.

dan-palmer (approved):

dan-palmer (approved):

PR #180, opened 2016-11-03
Use HTTPs locations for git repos for security reasons

lostie:
Solves issue flagged when running bundle:

```bash
The git source `git://github.com/shadabahmed/logstasher.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch
to the `https` protocol to keep your data secure.
```

**NOTE:** As for the git commit reference it needs to stay as is (for the moment), as there was no tagged version release after the reference commit used.


dan-palmer (approved):

dan-palmer:
@lostie I'd add that last note into the actual commit message so it's not lost

lostie:
Once I merge this, it will be recorded as part of the merged commit message.

PR #183, opened 2016-11-09
OS related ignores should be configured in the global git ignore file

lostie:
Self descriptive.

`~/.gitignore_global` should be updated in the developments machine, rather than making this changes in every single repo.


dan-palmer (approved):
Yeah hawww

PR #184, opened 2016-11-09
Remove seeds created directly through the controllers

lostie:
⚠️ **Unless there's a strong reason for this to be kept.**


dan-palmer (commented):
This is fine. User research et al aware this is being removed?

lostie:
> This is fine. User research et al aware this is being removed?

Ehrggg, not really. It shouldn't really matter, if for demo purposes, seeds are needed, they should be created some other way (that's my opinion anyway).

dan-palmer:
@lostie Just need to let everybody know that you're removing it and we will be seeding data differently in the future.

I agree that there are nicer ways of doing this.

dan-palmer (approved):

lostie:
Will be addressed in https://github.com/ministryofjustice/moving-people-safely/pull/212

PR #185, opened 2016-11-10
Simplify homepage controller

lostie:
- Makes routing more clear about the intent of the actions being performed
- Makes `params` used for the searches more clear about what they're actually used for
- Moves concerns of date picking to the `DatePickerRepresenter`


lostie:
This changes will be applied in https://github.com/ministryofjustice/moving-people-safely/issues/212

PR #186, opened 2016-11-11
Fix bug in 'Other' reason for move details

lostie:
[Trello](https://trello.com/c/HhYONuZc/338-bug-when-selecting-other-in-reason-for-move-in-move-details-page-it-shouldn-t-display-reason-details-above-give-details): When selecting 'Other' in Reason for move in Move details page it shouldn't display 'Reason details' above 'Give details'

### DIFF
--- a/app/models/forms/base.rb
+++ b/app/models/forms/base.rb
@@ -48,11 +48,11 @@ module Forms
         instance_methods.include?(method_name)
       end
 
-      def optional_field(field_name)
-        _define_attribute_is_on(field_name, TOGGLE_YES)
-        property(field_name, type: String)
+      def optional_field(field_name, options = {})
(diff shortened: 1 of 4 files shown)


app/models/forms/base.rb
```diff
@@ -65,6 +65,18 @@ def optional_details_field(field_name)
         reset attributes: ["#{field_name}_details"], if_falsey: field_name
       end
 
+      def property_with_details(field_name, options = {})
```
lostie, line 68: I've considered updating `optional_details_fields` instead, but I it wasn't very clear to me what the method name means, although I do know what is doing by looking at the code.

lostie:
This has now been approved by PM. Can one of you guys review and we'll merge if happy? 🍻

dan-palmer (approved):

PR #187, opened 2016-11-11
Improve arrow keys navigation on form fields

lostie:
[Trello](https://trello.com/c/NU4mH699/362-bug-check-behaviour-of-selecting-keyboard-arrows-when-radio-buttons-are-selected-in-application): This change enables the user to navigate between all the options available ('Yes', 'No' and 'unknown') with the use of arrow keys.

Previously by just hiding the 'unknown' ('Clear selection' label) option all together, the user would see the arrow keys navigation disappear as soon as it selected the 'Clear selection'.

This mocks the hiding by positioning the input field way off.

### DIFF
--- a/app/assets/stylesheets/moving_people_safely/_helper_classes.scss
+++ b/app/assets/stylesheets/moving_people_safely/_helper_classes.scss
@@ -62,6 +62,7 @@
 
 .mps-hide {
   @include js-only {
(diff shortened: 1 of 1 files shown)


dan-palmer (approved):
👍 Make sure the commentary about the PR gets captured in the merge.

PR #188, opened 2016-11-15
Make release date field optional on the offences form

lostie:
This PR solves [Trello ticket #320](https://trello.com/c/40Adobo5/320-blocker-make-release-date-on-offences-screen-and-case-reference-number-optional): 
* **release date** field is now optional
* current offence **case reference** is now optional

**NOTES:** 

- The `TextDate` field had to be coerced to **nilify** all blank values received in the form. This is not yet [supported](http://trailblazer.to/gems/disposable/api.html#nilify) by the current version of **Reform** used in the project and trying to upgrade it resulted in a lot of breaking changes which I decided to avoid for this ticket.
- This kind of validations `is_expected.to validate_presence_of` tend to be a bit dumb, as they only try to assert that a given method is being called appropriately, rather than testing the actual behaviour.


app/form_builders/mps_form_builder.rb
```diff
@@ -25,7 +25,7 @@ def radio_toggle_with_textarea(attribute, options = {})
 
   def field_without_label(field_type, attribute)
     content_tag :div,
-      class: form_group_classes(attribute),
+      class: form_group_classes(attribute.to_sym),
```
lostie, line 28: Until https://github.com/ministryofjustice/govuk_elements_form_builder/issues/68 is solved, this is required here to avoid unexpected errors.

Gemfile.lock
```diff
@@ -63,7 +63,7 @@ GEM
       minitest (~> 5.1)
       tzinfo (~> 1.1)
     addressable (2.4.0)
-    arel (7.1.2)
+    arel (7.1.4)
```
dan-palmer, line 66: Think this commit might be accidental :thinking:

dan-palmer (approved):

dan-palmer:
Splendid

PR #189, opened 2016-11-18
Remove considerations table from db schema

lostie:
The `considerations` table seems to have been added to the schema file by mistake.


lostie:
@dan-palmer can you review pls?

dan-palmer (approved):

PR #190, opened 2016-12-01
Pre-fill detainee details in new detainee form if available

lostie:
Implements [Trello ticket](https://trello.com/c/5LO9jkGJ/252-5-implement-person-interface-to-work-off-single-offender-id)

- Integrates MPS with the Offenders API so that in the new
detainees form the information is pre-filled if available in the
Offenders API (its using the NOMS ID)
- Adds a new service to retrieve data for a detainee from the Offenders
API given a NOMS id
- Sets up a configuration for Offenders API client based on the
applications environment
- Adds Index to noms_id in the detainees table to optimize searches
performed by this attribute
- Adds a safe guard to the detainee details controller to prevent multiple
detainees with the same NOMS id to be created
- Uses the existent redirect strategy when the detainee with the
provided NOMS id already exists
- Removes some of the unncessary complexity added by the misuse of Rails
flash


app/controllers/detainee_details_controller.rb
```diff
@@ -44,17 +34,41 @@ def update
 
   private
 
-  def redirect_to_move_or_profile
+  def redirect_if_detainee_exists
```
dan-palmer, line 37: How about:

```ruby
def redirect_if_detainee_exists
  if params[:prison_number].present? && Detainee.exists?(prison_number: params[:prison_number])
    redirect_to_move_or_profile(flash: { warning: t('alerts.detainee_already_exists') })
  end
end
```

app/controllers/detainee_details_controller.rb
```diff
@@ -44,17 +34,41 @@ def update
...
+      redirect_to new_move_path(detainee), options
     end
   end
 
   def form
     @_form ||= Forms::DetaineeDetails.new(detainee)
   end
+
+  def detainee_default_attrs
+    return {} unless params[:prison_number].present?
```
dan-palmer, line 59: We can switch the if not's to if's using `blank?` - think this applies to a few places.

db/migrate/20161201120724_add_index_to_prison_number_on_detainees_table.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToPrisonNumberOnDetaineesTable < ActiveRecord::Migration[5.0]
+  def change
+    add_index :detainees, :prison_number
```
dan-palmer, line 3: 👍

app/services/detainee_details_fetcher.rb
```diff
@@ -0,0 +1,64 @@
...
+    present_names = names.select(&:present?)
+    return if present_names.empty?
+    present_names.join(' ')
+  end
+
+  def map_gender(gender)
+    return unless gender.present?
+    return 'male' if gender.casecmp('M').zero?
+    return 'female' if gender.casecmp('F').zero?
+    nil
```
dan-palmer: How about:
```ruby
{ 'm' => 'male', 'f' => 'female' }[gender.downcase] if gender.present?
```

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 ruby '2.3.0'
 
 gem 'activerecord-session_store'
+gem 'countries', git: 'https://github.com/hexorx/countries.git'
```
dan-palmer: Does it need the git option?

app/services/detainee_details_fetcher.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def map_nationality(nationality_code)
+    return unless nationality_code.present?
+    country = ISO3166::Country.new(nationality_code)
+    return nationality_code unless country
+    country.nationality
```
dan-palmer:
```ruby
def country(nationality_code)
  country = ISO3166::Country.new(nationality_code)
  (country && country.nationality) || nationality_code
end

File: spec/services/detainee_details_fetcher_spec.rb [hunk 7]
```diff
@@ -0,0 +1,232 @@
+require 'rails_helper'
+
+RSpec.describe DetaineeDetailsFetcher do
```
dan-palmer, line 3: Cool tests bro

app/controllers/detainee_details_controller.rb
lostie: Unfortunately, you need to assign the `@_detainee` variable in order to update this very same variable which is set in the in the controller that detainee_details_controller inherits from.

dan-palmer: Ah.. I did wonder that. Fair enough 😸

lostie:
@dan-palmer ♻️ (ed)

PR #194, opened 2016-12-12
SSO integration

lostie:
Trello [ticket](https://trello.com/c/0L7qDexa/323-5-single-sign-on-integration)

This feature includes:
- Adding a link to the login page to enable **SSO** through the existent MOJ SSO application
- Adds configurations to setup the MOJ SSO integration for the different environments
- Assumes **no** multiple SSO providers are supported, only one at a time


lostie:
@alan Could you please review?

app/models/user.rb
```diff
@@ -1,4 +1,16 @@
...
   devise :database_authenticatable, :recoverable,
-    :trackable, :validatable, :timeoutable
+    :trackable, :validatable, :timeoutable,
+    :omniauthable, omniauth_providers: [:mojsso]
+
+  def self.from_omniauth(auth)
+    where(provider: auth.provider, uid: auth.uid).first_or_create do |user|
+      user.provider = auth.provider
+      user.uid = auth.uid
+      user.email = auth.info.email
```
alan: Have you got a unique index on email? If so you could have the case that someone uses sso after having an account which will result in an error.

lostie: Yes, there's a unique index for the email:

```sql
Indexes:
    "users_pkey" PRIMARY KEY, btree (id)
    "index_users_on_email" UNIQUE, btree (email)
```

What's the recommended strategy to deal with this cases? Check for user by email first and update only the `provider` and `uid`?

alan: Depends on your use case, in PVB `moj-sso` is the only way to sign in so we don't have this scenario. I don't know what is the use case to store the `(provider, uid)` on user...

If you can, making sso the only way to sign in to your app would be the simplest thing to do. For your users (they would only have one account in sso to login) and in the code (you don't even need devise as a dependency).

lostie:
@alan ♻️  with suggested improvements.

PR #200, opened 2016-12-22
Security risk changes

lostie:
[Trello #350](https://trello.com/c/dyXCb5M5/350-5-as-someone-working-in-security-in-a-prison-when-filling-in-a-digital-per-i-want-to-be-able-to-complete-security-details-as-par)

 - Removes unnecessary questions for security risk section that are no
    longer used
- Adds new columns for new questions required by the security risk
    section
 - Change form for security risk section to comply with the new
    requirements

As an added feature of this PR, it support a more flexible way of localising the data of the application, which was a bit limited by the restrictions that the use of the gov-form-builder gem sets on the application.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,8 +37,15 @@ Metrics/AbcSize:
   Exclude:
     - 'spec/**/*'
 
+Metrics/ClassLength:
+  Exclude:
+    - 'app/form_builders/mps_form_builder.rb'
+
 Metrics/MethodLength:
   Max: 15
+  Exclude:
+    - 'app/form_builders/mps_form_builder.rb'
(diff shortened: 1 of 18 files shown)


.rubocop.yml
```diff
@@ -37,8 +37,15 @@ Metrics/AbcSize:
   Exclude:
     - 'spec/**/*'
 
+Metrics/ClassLength:
+  Exclude:
+    - 'app/form_builders/mps_form_builder.rb'
```
lostie, line 42: I've left a note in the first commit explaining why I decided to exclude this file from Rubocop.

cesidio (approved):

PR #213, opened 2017-01-12
Change radios to checkboxes for risk trafficking options

lostie:
[Trello #424](https://trello.com/c/yKOLdie4/424-risk-of-trafficking-drugs-and-alcohol-question-interaction-amend)

The requirements is for the user to be able to select either or both so
it makes sense the substance misuse section follows the same approach as
other risk sections, which is to display all the possible values as
checkboxes and the user can select all the ones that apply.


lostie:
@himal-mandalia Could you please 👁‍🗨 👁‍🗨 ?

himal-mandalia (approved):

PR #232, opened 2017-02-14
Update detainee information present in move printout

lostie:
[Trello #21](https://trello.com/c/xQuAIRU9/21-5-as-someone-printing-out-a-completed-digital-per-i-want-to-see-the-detainee-s-personal-details-displayed-on-the-summary-sheet-d)

These changes include:
- Update information displayed in the move printout related with the
detainee according to new requirements.
- Minor refactoring in the print routes. They now fall under the moves
namespace to be very clear what the print is all about.
- Add a controller to server detainee images. THe background for this is
the fact that the images are being stored as a blob in the database and
trying to directly render them in the src attribute of an image DOM tag
has some issues, specially in IE8, where there's a limitation in size
for the image data being rendered. So to solve this problem overall, the
application serves as an endpoint the image associated to the detainee.
- Add a detainee presenter to the move printout to encapsulate all the
printout logic related with the detainee.


app/presenters/print/detainee_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+    def cro_number
+      model.cro_number.present? ? model.cro_number : 'None'
+    end
+
+    def pnc_number
+      model.pnc_number.present? ? model.pnc_number : 'None'
+    end
+
+    def date_of_birth
+      model.date_of_birth.strftime('%d %b %Y')
```
dan-palmer: Just spotted a `humanized` date format matching this. We should probably use that for consistency.

spec/models/move_spec.rb
```diff
@@ -1,7 +1,33 @@
...
 
 RSpec.describe Move, type: :model do
-  subject { described_class.new }
+  subject(:move) { described_class.new }
+
+  describe '#issued?' do
+    context 'when the move is still in not started status' do
+      let(:move_workflow) { FactoryGirl.build(:move_workflow, status: :not_started) }
+      let(:move) { FactoryGirl.build(:move, move_workflow: move_workflow) }
+      specify { expect(move.issued?).to eq(false) }
```
dan-palmer: we could do:
```ruby
it { is_expected.not_to be_issued }
```

dan-palmer (approved):
🚓

PR #235, opened 2017-02-16
Spring cleaning

dan-palmer:
- [x] Remove ruby version from Gemfile
- [x] Remove unnecessary ENV vars from asset command


lostie (approved):
LGTM 👍

PR #236, opened 2017-02-20
Update move information present in move printout

lostie:
[Trello #14](https://trello.com/c/uVw8q4on/14-2-as-someone-printing-out-a-completed-per-i-want-to-see-a-summary-of-the-move-details-planned-for-the-detainee)

These changes include:
- Update information displayed in the move printout related with the
move according to new requirements.
- Add new assertions to the printing feature test to cover the new
changes in the printout

**NOTE:** It appears Chrome print preview does not respect the style
specified in the print.scss. After some investigation without much
success, I've added some specific CSS for media print which seems to
solve this issue. Even though it turns out to repeat some of the CSS
styles, I will leave it as is until a better DRY solution is found.

### DIFF
--- a/app/views/moves/print/show.html.slim
+++ b/app/views/moves/print/show.html.slim
@@ -77,17 +77,18 @@ html
                 = image_tag "photo_unavailable.png", size: '128x180'
       div.grid.move
         div.grid-row.header-row.small = 'Move details'
-        div.grid-row.padded
+        div.grid-row.padded.padded-bottom
           div.grid-column.column-one-third
             div.title Date of travel
-            div.strong-text = move.date
+            div.strong-text = move.date.to_s(:humanized)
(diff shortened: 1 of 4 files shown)


app/views/moves/print/show.html.slim
```diff
@@ -77,17 +77,18 @@ html
                 = image_tag "photo_unavailable.png", size: '128x180'
       div.grid.move
         div.grid-row.header-row.small = 'Move details'
-        div.grid-row.padded
+        div.grid-row.padded.padded-bottom
           div.grid-column.column-one-third
             div.title Date of travel
-            div.strong-text = move.date
+            div.strong-text = move.date.to_s(:humanized)
```
dan-palmer, line 83: 👁

dan-palmer (approved):

PR #240, opened 2017-02-24
Display missing text on summary if unanswered

dan-palmer:
Prior to this fix it was possible to visit a summary page and see “No”’s
for questions that do not have their top level question answered.


lostie (approved):
🎩

PR #241, opened 2017-02-27
Print out risk section

lostie:
[Trello #31](https://trello.com/c/j1XIqWpc/31-5-as-someone-printing-out-a-completed-per-i-want-to-see-a-summary-of-all-the-detainee-s-risk-information)

Add risk assessment content to the move printout.

Also:
- Add new print presenters for risk sections and subsections
- Add new print presenter for an assessment question
- Tweak the way pages break in the print media

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -51,7 +51,7 @@ $is-print: true;
     width: (100% / 6);
   }
 
-  .section-table {
+  .section-sub-table {
     page-break-inside: avoid;
   }
 }
@@ -119,8 +119,16 @@ table {
   width: (100% / 6);
 }
 
+.column-20-percent {
+  width: 20%;
+}
+
 .column-30-percent {
   width: 30%;
+}
+
+.column-50-percent {
+  width: 50%;
 }
 
 .column-70-percent {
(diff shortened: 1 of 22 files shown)


app/assets/stylesheets/print.scss
```diff
@@ -51,7 +51,11 @@ $is-print: true;
     width: (100% / 6);
   }
 
-  .section-table {
+  // .section-table {
```
dan-palmer: ♻️

app/models/sections/base_section.rb
```diff
@@ -1,4 +1,8 @@
 class BaseSection
+  def name
+    raise NotImplementedError
```
dan-palmer: `NotImplementedError` is recommended for use when a feature is not implemented on the current platform - so if the current OS or Ruby version doesn't support something.

spec/presenters/print/assessment_question_presenter_spec.rb
```diff
@@ -0,0 +1,329 @@
...
+
+  subject(:presenter) { described_class.new(question, assessment, section) }
+
+  before do
+    allow(assessment).to receive(question).and_return(answer)
+  end
+
+  describe '#label' do
+    before do
+      I18n.backend.store_translations(:en, {
```
dan-palmer: I like this I18n testing approach

dan-palmer: Saying that we could have a helper to make it less verbose

dan-palmer (approved):

PR #243, opened 2017-03-06
Pdf documents

dan-palmer:


app/services/pdf_generator.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def move_presenter
+    @move_presenter ||= Print::MovePresenter.new(move)
+  end
+
+  def detainee_presenter
+    @detainee_presenter ||= Print::DetaineePresenter.new(detainee)
+  end
+
+  def move
```
lostie: Can use a private `attr_reader :move`

app/services/pdf_generator.rb
```diff
@@ -0,0 +1,42 @@
...
+  def move_presenter
+    @move_presenter ||= Print::MovePresenter.new(move)
+  end
+
+  def detainee_presenter
+    @detainee_presenter ||= Print::DetaineePresenter.new(detainee)
+  end
+
+  attr_reader :move
+  private :move
```
lostie: 🔪

app/services/pdf_generator.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def detainee
+    move.detainee
+  end
+
+  def risk
```
lostie: 💡 **[no blocker]** Could make use of the delegator for these methods:

```ruby
delegate :risk, :healthcare, to: :detainee
```

lostie (approved):
LGTM 👍

PR #246, opened 2017-03-15
Ensure prison number is always upcased

dan-palmer:
The prison number should be upcased for querying the NOMIS API and
storing in the DB. A user doesn’t always input the prison number in
its upcased form which can result in failed search results on the
NOMIS API.

### DIFF
--- a/app/models/forms/detainee.rb
+++ b/app/models/forms/detainee.rb
@@ -26,6 +26,10 @@ module Forms
       errors.add(:date_of_birth) unless date_of_birth.is_a? Date
     end
 
+    def prison_number=(value)
+      value && super(value.upcase)
+    end
+
     def genders
       GENDERS
     end
--- a/app/models/forms/search.rb
+++ b/app/models/forms/search.rb
@@ -16,6 +16,10 @@ module Forms
       presence: true,
       format: { with: PRISON_NUMBER_REGEX }
 
+    def prison_number=(value)
+      value && super(value.upcase)
+    end
+
     def detainee
       @_detainee ||= ::Detainee.find_by(_at[:prison_number].matches(prison_number)) if valid?
     end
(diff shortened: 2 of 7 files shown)


lostie:
If you make a direct request to `http://<base_url>/detainees/new?prison_number=A1418At` you will still search for the prison number as it was inputted by the user:

```ruby
Started GET "/detainees/new?prison_number=A1418At" for 127.0.0.1 at 2017-03-20 10:32:46 +0000
Processing by DetaineesController#new as HTML
  Parameters: {"prison_number"=>"A1418At"}
...
[Detainees::DetailsFetcher] NOMIS API: Requesting details for offender with NOMS id A1418At
[Detainees::DetailsFetcher] NOMIS API: #<Nomis::HttpError: Unexpected status 404 calling GET /nomisapi/offenders/A1418At: {"error"=>{"message"=>"Offender Not Found"}}>
```

Looking into it...

lostie:
@dan-palmer I've updated your PR to ensure the `prison_number` used in the _forms/API/database_ is consistent. Can you 🙏 review? 🎩

dan-palmer:
@lostie thanks dude

dan-palmer:
LGTM 🚀

app/models/forms/detainee.rb
```diff
@@ -26,6 +26,10 @@ def validate_date_of_birth
       errors.add(:date_of_birth) unless date_of_birth.is_a? Date
     end
 
+    def prison_number=(value)
+      value && super(value.upcase)
```
dan-palmer, line 30: Seems as this in a few places now prison number could become a value object. Something like:

```ruby
class PrisonNumber < SimpleDelegator
  def initialize(value)
    super(value&.upcase)
  end
end
```

PR #248, opened 2017-03-20
Sso only

dan-palmer:
- [x]  remove devise log in screens
- [x]  remove old fields on user table
~~- [ ]  tidy up sign in and login spec helper methods~~


app/controllers/detainees/images_controller.rb
```diff
@@ -1,6 +1,5 @@
 module Detainees
   class ImagesController < ApplicationController
-    respond_to :jpg
```
lostie, line 3: 👺 : What's the reason for this change (specially in this PR)  ?

lostie (commented):
For development purposes, these changes imply that we will need a local version of the MOJ SSO running.

It's likely that the instructions to setup the application need to be updated to:
- Have MOJ SSO app as an **explicit requirement** to run the app
- Update instructions to setup MOJ SSO app so that, ideally, the user only needs to run a _single_ command that sets the application up and running (if it's a dockerized version or not, it's up for discussion)
- Instructions about the default user that can be used to login into the app

app/controllers/detainees/images_controller.rb
dan-palmer: Should be in the commit message for that change. The respond_to method was only available because of Devise adding in the responders gem.

app/models/user.rb
```diff
@@ -22,4 +22,8 @@ def first_or_create_by_uid_and_provider(auth)
       end
     end
   end
+
+  def full_name
+    [first_name, last_name].select(&:present?).join(' ')
```
dan-palmer: A faster way to do this(not that that matters) but with a lil less code is:
```ruby
def full_name
  "#{first_name} #{last_name}".strip
end
```

lostie:
@dan-palmer I've added an isolated [commit](https://github.com/ministryofjustice/moving-people-safely/pull/248/commits/2e797f2736db3c1e6707cbb4dcb776e2abcea36a) that addresses the issue of logging out in _mock mode_ (using the logout URL contained in the SSO data).

Basically, the approach I took was to redirect the user to the _new_session_path_ in development, and to create a (very basic) mocked SSO app for testing purposes.

Please 👓  ;)

app/models/sso/identity.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      def from_session(hash)
+        user = User.find(hash.fetch('user_id'))
+        options = {
+          profile_url: hash.fetch('profile_url'),
+          logout_url: hash.fetch('logout_url'),
+          permissions: hash.fetch('permissions')
+        }
+        new(user, options)
+      rescue KeyError
```
dan-palmer: Don't think this is necessary. 🛡 ⚔️ 

It would suggest we've allowed bad data into the system to begin with(as its session retrieval).

For the `options` we could replace the fetches with a `slice`.

app/models/sso/identity.rb
```diff
@@ -0,0 +1,62 @@
...
+  class Identity
+    InvalidData = Class.new(StandardError)
+
+    class << self
+      def from_omniauth(auth_hash)
+        info = auth_hash.fetch('info')
+        user = User.from_omniauth(auth_hash)
+        options = extract_additional_info(info)
+        new(user, options)
+      rescue
```
dan-palmer, line 9: This seems to be an exceptional situation

app/models/user.rb
dan-palmer: bump

config/initializers/devise.rb
```diff
@@ -1,317 +0,0 @@
...
-
-  # Email regex used to validate email formats. It simply asserts that
-  # one (and only one) @ exists in the given string. This is mainly
-  # to give user feedback and not to assert the e-mail validity.
-  # config.email_regexp = /\A[^@]+@[^@]+\z/
-
-  # ==> Configuration for :timeoutable
-  # The time you want to timeout the user session without activity. After this
-  # time the user will be asked for credentials again. Default is 30 minutes.
-  config.timeout_in = 60.minutes
```
dan-palmer, line 209: Devise provided a timeout. Should we add this to the `session_store`?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def auth_hash
+    @auth_hash ||= request.env['omniauth.auth']
+  end
+
+  def sso_path
+    '/auth/mojsso'
+  end
+  helper_method :sso_path
+
+  def build_url(url, redirect_url = nil)
```
dan-palmer: Think this shouldn't be generic & should just be `def logout_url` with no args.

app/controllers/application_controller.rb
```diff
@@ -24,4 +24,30 @@ def redirect_unless_document_editable
...
+  end
+  helper_method :user_signed_in?
+
+  def sso_identity
+    @sso_identity ||= set_sso_identity
+  end
+
+  def set_sso_identity
+    session[:sso_data] && SSO::Identity.from_session(session[:sso_data])
+  rescue SSO::Identity::InvalidData
```
dan-palmer: Do we have a test for this?

app/assets/stylesheets/moving_people_safely/_authentication.scss
```diff
@@ -1,9 +1,12 @@
-.sign-out {
+#signed-in-box {
```
dan-palmer, line 1: OMG 🍼 🍺

lostie:
👍

dan-palmer:
Beautiful 👸 🤴

PR #250, opened 2017-03-23
Add new option to display flexible toggle content for alert in move printout

lostie:
In the alerts section for the printout is now possible to specify a new
option that enable a specifi alert to have a specific content for the
toggle alert instead of the default red tick that is displayed when the
alert is on.

Relates to: [Trello #30](https://trello.com/c/YbFEfMIj/30-as-someone-printing-out-a-completed-per-i-want-to-see-all-the-detainee-s-risk-markers)

### DIFF
--- a/app/presenters/print/alert_presenter.rb
+++ b/app/presenters/print/alert_presenter.rb
@@ -7,12 +7,17 @@ module Print
       @view_context = options.fetch(:view_context)
       @status = options.fetch(:status, :off)
       @title = options[:title]
+      @toggle = options[:toggle]
       @text = options[:text]
     end
 
     def title
       return @title if @title.present?
       h.t("print.alerts.title.#{name}", default: name.to_s.humanize)
+    end
+
+    def on?
+      status == :on
     end
 
     def to_s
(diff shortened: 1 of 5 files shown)


spec/presenters/print/alert_presenter_spec.rb
```diff
@@ -36,13 +36,34 @@
     end
   end
 
+  describe '#on?' do
```
dan-palmer: Should `on?` be public? Looks like a private method.

app/presenters/print/alert_presenter.rb
```diff
@@ -41,6 +46,18 @@ def build_text_content
...
+    def build_toggle_content
+      if toggle.present?
+        h.content_tag(:span, toggle, class: 'alert-toggle')
+      elsif on?
+        h.wicked_pdf_image_tag('ic_red_tick.png')
+      end
+    end
+
     def alert_class
       status == :on ? 'alert-on' : 'alert-off'
```
dan-palmer: We could use the new `on?` method here too.

app/presenters/print/alert_presenter.rb
```diff
@@ -26,13 +31,13 @@ def to_s
...
 
-    attr_reader :view_context
+    attr_reader :view_context, :toggle
     alias h view_context
 
     def build_status_content
       h.content_tag(:div, class: "image #{alert_class}") do
         alert_contents = [h.content_tag(:span, title, class: 'alert-title')]
-        alert_contents << h.wicked_pdf_image_tag('ic_red_tick.png') if status == :on
+        alert_contents << toggle_content if toggle_content
```
dan-palmer: Think we can just ignore the `if`. If that call returns nil it wont affect `safe_join`.

lostie:
♻️

dan-palmer (approved):

PR #251, opened 2017-03-23
Pdf enhancements

dan-palmer:
Padding on the subsections, fix the page size(reduce the zoom effect), change the pdf testing approach.


app/assets/stylesheets/print.scss
```diff
@@ -319,99 +318,6 @@ table.vehicles {
   }
 }
 
-table.detainee {
```
lostie, line 327: 🌈 👯 🌈

spec/features/printing_spec.rb
```diff
@@ -115,89 +81,13 @@
       })
     }
 
-    let(:front_cover_alerts){
```
lostie, line 118: image

lostie (approved):
LGTM 👍

PR #256, opened 2017-03-28
Remove past offences

dan-palmer:
- [x] Update current offences UI according to new [design](https://dsdmoj.atlassian.net/wiki/pages/viewpage.action?spaceKey=MPS&title=Offences+1.X&preview=/137560281/141236461/Offences%20-%20complete.png)
- [x] Remove from UI
- [x] Remove from PDF

#### "Profile" Page:
screen shot 2017-03-28 at 15 44 18

#### Offences Page:
screen shot 2017-03-28 at 15 45 16


dan-palmer:
@lostie The modelling of offences -> current_offences is weird. The ❓ is do we sort it now or do we wait for the model refactor and see what happens?

app/controllers/offences_controller.rb
```diff
@@ -25,10 +25,6 @@ def add_offence
       form.deserialize form_data
       form.add_current_offence
       render :show, locals: { form: form }
-    elsif params.key? 'offences_add_past_offence'
```
lostie, line 28: image

:trollface:

lostie (approved):
LGTM 👍 

Regarding tacking the offences/current offences modelling I think they should be tackled. If in this PR or as a separate task, I would probably go with a separate one.

PR #258, opened 2017-03-30
Exclude pdf spec from CI

dan-palmer:
**This exclusion is temporary**

Differences in the environment are causing this to fail.We are
considering running the specs in a container or changing the way we test
the PDF by testing the HTML we pass to Wicked.


lostie (approved):
👍

PR #259, opened 2017-03-31
Add headers to PDF printout pages

lostie:
**Note:** The tool to generate the PDF is a bit flakey. Any minor changes to
the layout seem to have a major impact in the display, some even making
the PDF to fail to render.

There's some known display issues, maybe directly or indirectly related with
the addition of the headers.


app/views/moves/print/header.html.slim
```diff
@@ -0,0 +1,35 @@
+doctype html
+
+javascript:
+  function number_pages() {
+    var vars={};
+    var x=document.location.search.substring(1).split('&');
+    for(var i in x) {var z=x[i].split('=',2);vars[z[0]] = decodeURIComponent(z[1]);}
+    var x=['frompage','topage','page'];
```
dan-palmer, line 8: Is `frompage` necessary?

app/views/moves/print/header.html.slim
```diff
@@ -0,0 +1,35 @@
+doctype html
+
+javascript:
+  function number_pages() {
+    var vars={};
+    var x=document.location.search.substring(1).split('&');
```
dan-palmer, line 6: We should probably gives these `var`s better names.

app/views/moves/print/header.html.slim
```diff
@@ -0,0 +1,35 @@
...
+    tr style="font-family: sans-serif"
+      td
+        span style="font-weight: bold"
+          == detainee.identifier
+        span
+          ==< detainee.forenames
+      td style="text-align: right; padding-right: 5px;"
+        div
+          | Page
+          span.page ==<> ''
```
dan-palmer, line 33:
```brainfuck
++++++++++[>+++++++>++++++++++>+++>+<<<<-]>++.>+.+++++++..+++.>++.<<+++++++++++++++.>.+++.------.--------.>+.>.
```

dan-palmer:
👍

PR #260, opened 2017-03-31
Adds a rake tasks to soft delete all unissued escorts past their date

lostie:
Includes:
- Migration to add deleted_at field to moves
- Add default scope to moves to only return non-deleted moves by default
- Add service to delete all unissued moves past their date
- Add rake task to call the service

### DIFF
--- a/app/models/escort.rb
+++ b/app/models/escort.rb
@@ -3,10 +3,13 @@ class Escort < ApplicationRecord
   has_one :detainee
   has_one :move
 
+  default_scope { where(deleted_at: nil) }
+
   scope :for_date, ->(date) { eager_load(move: [:move_workflow]).where(moves: { date: date }) }
   scope :with_incomplete_risk, -> { joins(move: [:risk_workflow]).merge(Workflow.not_confirmed) }
   scope :with_incomplete_healthcare, -> { joins(move: [:healthcare_workflow]).merge(Workflow.not_confirmed) }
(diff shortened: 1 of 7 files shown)


dan-palmer (approved):
⚒ 🚢

app/services/moves/delete_historic_unissued.rb
```diff
@@ -0,0 +1,11 @@
+module Moves
+  class DeleteHistoricUnissued
+    def self.call
+      new.call
+    end
+
+    def call
+      Move.active.where('date(now()) - date(date) >= 1').in_batches.update_all(deleted_at: Time.now.utc)
```
dan-palmer: What do you think about having this scope on the model?
```ruby
scope :unissued, -> { active.where('date(now()) - date(date) >= 1') }
```

lostie: as the [current code is](https://github.com/ministryofjustice/moving-people-safely/blob/42ddd2862d154d835af18f15ffb2eca2289b83d1/app/models/move.rb#L19),  just turns out to be more confusing (at least the naming) IMO.

Also `date(now()) - date(date) >= 1` is kind of a filter. You can have *unissued* moves that are still active, as in, their date has not been reached, so they haven't yet been printed.

If we find a clear name for this, I'm happy to move this to a scope of its own.

dan-palmer:
Tested on staging :shipit:

PR #261, opened 2017-04-03
Nomis offences

dan-palmer:
- [x] Fetch nomis offences from nomis endpoint
- [x] Extract current offences only
- [x] Update the UI with the fetched offences
- [x] Inform the user when the api errors via the flash

Ticket: [#87](https://trello.com/c/RTRTgXge/87-5-implement-offences-interface-to-work-off-nomis)

screen shot 2017-04-03 at 17 38 16

### DIFF
--- a/app/services/detainees/base_fetcher.rb
+++ b/app/services/detainees/base_fetcher.rb
@@ -29,6 +29,20 @@ module Detainees
       }[http_status] || 'api_error'
     end
 
+    def with_error_handling(&_blk)
+      yield
+    rescue Nomis::HttpError => e
+      log_api_error(e.inspect)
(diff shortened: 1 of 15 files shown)


app/services/detainees/base_fetcher.rb
```diff
@@ -22,13 +22,31 @@ def error_response(error_code)
...
+    end
+
     def error_code_for_http_status(http_status)
       {
         404 => 'not_found',
         400 => 'invalid_input'
       }[http_status] || 'api_error'
     end
 
+    def with_error_logging(&_blk)
```
lostie: 💭 Better wording for this would be `with_error_handling`

lostie (approved):
LGTM 👍

PR #263, opened 2017-04-05
Hide remove checkbox for multiples

dan-palmer:
Same approach as before but handles wider screens.

#### Before:
screen shot 2017-04-05 at 14 50 18

#### After:
screen shot 2017-04-05 at 14 51 13


lostie (approved):
🚢

PR #264, opened 2017-04-07
Escort as top level object

lostie:
Refactor data modelling to guarantee idempotency of escort

As it stands every time a new escort was created the information used
for a previous escort, namely, detainee information, risk assessment,
healthcare assessment, offences would be re-used and most likely
modified which meant we would track of the information used to perform
the previous escort.

These changes ensure that all the information used in an escort are
related only to that escort and cannot be modify outside the scope of
the escort itself.

This turns out to be quite a big change in terms of:
- Routes - all the data required to be created/modified is now under the
escorts scope
- Detainee and Move are not longer directly related but instead they
    belong to an Escort record. There's potential, in the future for these
    two models to be normalised and all the information related with them to
    be present directly in the Escort record. Due to the nature of the
    refactor we have refrained from doing any more changes than the MVP.
- The risk and healthcare assessments and the offences are currently
    related to a detainee, so the escort record access its information by
    delegating their call to the detainee. There's potential for this to be
    move directly to the escort, so that the escort has a directly
    relationship with the data it requires
- The workflows associated with the move, the risk and healthcare
    assessments, and the offences, are defined in a very convoluted way
    where the detainee holds the relation to the data but the move holds the
    relation to the status of the data. In this PR we masked this by having
    the models themselves to response to status enquires. Under the hood, we
    are still navigation to this convoluted way of accessing the status. In
    a future iteration we will simplify the process by having the model to
    respond directly to this status enquires.
- This PR starts to establish the responsability to the status of the
    escort to the escort itself, although it currently does it by delegating
    some of its status checks to the Move record. This is a transiction step
    as the Move record has a lot of responsability that shouldn't really
    have, but again, refactoring this would increase even more the size of
    the PR which is already quite big.

Future iterations:
- Establish correct relationships between models
- Have escort to control directly its status


app/views/escorts/show.html.slim
```diff
@@ -0,0 +1,21 @@
+= content_for :breadcrumbs do
+  - breadcrumbs_for_page root: true do
+    - breadcrumb detainee_breadcrumb(escort.detainee), '#'
+
+.escort
+  = render partial: 'header', locals: { detainee: escort.detainee, alerts: DetaineeProfileAlertsPresenter.new(escort.detainee) }
```
lostie: We could rename this to `EscortAlertsPresenter`

db/migrate/20170405094953_re_create_escorts.rb
```diff
@@ -0,0 +1,9 @@
+class ReCreateEscorts < ActiveRecord::Migration[5.0]
+  def change
+    create_table :escorts, id: :uuid do |t|
+      t.string :prison_number
```
dan-palmer, line 4: Should we remove the prison number from the detainee table and just delegate to the escorts prison number for now?

app/services/escort_creator.rb
```diff
@@ -0,0 +1,43 @@
...
+  private
+
+  attr_reader :prison_number
+
+  INCLUDE_GRAPH = [
+    { detainee: [
+      :risk,
+      { healthcare: [:medications] },
+      { offences: [:current_offences] }
+] },
```
dan-palmer: 🚓 such padding

app/controllers/application_controller.rb
```diff
@@ -5,23 +5,11 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!
 
-  helper_method :offences,
```
dan-palmer, line 8: 👍

app/views/homepage/_no_result.html.slim
```diff
@@ -1,6 +1,9 @@
 p
-  | There is no previously created profile on
+  | There is no previously PER record on
```
dan-palmer: Probably should read `There are no previously created PER records on` ..

spec/features/search_spec.rb
```diff
@@ -61,16 +67,17 @@ def expect_error_message
...
+    expect(page).to have_content(detainee.prison_number).
+      and have_content(detainee.surname).
+      and have_content(detainee.date_of_birth.strftime('%d %b %Y')).
       and have_content(move.to).
       and have_content(move.date.strftime('%d %b %Y'))
   end
 
   def expect_result_with_no_move(detainee)
-    expect(page).to have_link('Add new move', href: new_detainee_move_path(detainee))
+    # expect(page).to have_button('Add new move', href: new_escort_move_path(detainee.escort))
```
dan-palmer: ♻️

spec/services/create_escort_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def expect_offences_to_be_cloned(existent_escort, new_escort)
+    offences_attributes = existent_escort.offences.attributes.except(*except_offences_attributes)
+    expect(new_escort.offences.id).not_to eq(existent_escort.offences.id)
+    expect(new_escort.offences).to have_attributes(offences_attributes)
+    expect(new_escort.offences.status).to eq('needs_review')
+
+    expected_current_offences_attributes = existent_escort.offences.current_offences.map { |co| co.attributes.except(*except_current_offences_attributes) }
+    current_offences_attributes = new_escort.offences.current_offences.map { |co| co.attributes.except(*except_current_offences_attributes) }
+    expected_current_offences_attributes.each do |expected_attributes|
```
dan-palmer: Given the hashes have the same schemas on the lhs and rhs of the test would `match_array` not work here? It doesn't take into account the order of the arrays just that they contain the same things.

```ruby
expect(current_offences_attributes).to match_array(expected_current_offences_attributes)
```

dan-palmer (approved):

spec/services/create_escort_spec.rb
lostie: That does not work as expected:

```ruby
     Failure/Error: current_offences_attributes.to match_array(expected_current_offences_attributes)

     NoMethodError:
       undefined method `>=' for #<RSpec::Matchers::BuiltIn::ContainExactly:0x007fb4c6ceff68>
     # ./spec/services/create_escort_spec.rb:70:in `expect_offences_to_be_cloned'
     # ./spec/services/create_escort_spec.rb:31:in `block (3 levels) in <top (required)>'
```

The current test runs as expected, so leaving it as is for now. Can be re-visited if some other way is found to achieve the same expectation.

lostie:
♻️

spec/services/create_escort_spec.rb
dan-palmer: Looks like a typo. The `current_offences_attributes` needs to be passed to the expect method in order to return an object that responds to `to`.

PR #265, opened 2017-04-11
Simplify offences

dan-palmer:
Remove the offences model that sits between the `current_offences` and the now defunct `past_offences`. Then use a common language across the app, in this case we want to refer to any offences as that just offences and not current offences.

Some of the copy has been kept as current offences this is due to the fact that we are under going a  copy review at present and it might add a bit of confusion if that changed.

Continuation of ticket: [#87](https://trello.com/c/RTRTgXge/87-5-implement-offences-interface-to-work-off-nomis)


dan-palmer:
⏰ Need to rebase on to #264

dan-palmer:
Resolved by #270

PR #266, opened 2017-04-11
Add ability to retrieve application version info

dan-palmer:
It’s useful across the environments to be able to see what has been
deployed and when it was deployed.

The information returned is consistent with what other apps in the
department return.

Example response from dev:

```json
{
  "build_date":"2017-04-11T14:31:18+0000",
  "build_tag":"pingendpoint.9a6ab35",
  "commit":"9a6ab3526cbf4b26869be7042398682b6a40b241"
}
```

### DIFF
--- a/app/controllers/ping_controller.rb
+++ b/app/controllers/ping_controller.rb
@@ -1,0 +1,17 @@
+class PingController < ActionController::Base
+  def show
+    render json: version_info
+  end
+
+  private
+
+  def version_info
+    {
+      build_date: ENV.fetch('APP_BUILD_DATE'),
+      build_tag: ENV.fetch('APP_BUILD_TAG'),
+      commit: ENV.fetch('APP_GIT_COMMIT')
+    }
+  rescue KeyError
+    {}
+  end
+end
--- a/spec/requests/ping_spec.rb
+++ b/spec/requests/ping_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe 'Pings', type: :request do
+  describe 'GET /ping' do
+    context 'when the correct environment variables have been set during the deploy process' do
+      before do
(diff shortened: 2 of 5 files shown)


spec/requests/ping_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe 'Pings', type: :request do
+  describe 'GET /ping' do
+    context 'when the correct environment variables have been set during the deploy process' do
+      before do
+        allow(ENV).to receive(:fetch).with('APP_BUILD_DATE').and_return('2017-04-06T14:47:13+0000')
```
lostie: **[not a blocker]** 💡 I've come across this [gem](https://github.com/thoughtbot/climate_control) that achieves the same thing in a slightly cleaner and more [controlled](https://github.com/thoughtbot/climate_control#why-use-climate-control) away.

For this specific case it can be used with a _RSpec around_ block

app/controllers/ping_controller.rb
```diff
@@ -0,0 +1,17 @@
+class PingController < ActionController::Base
+  def show
+    render json: version_info
+  end
+
+  private
+
+  def version_info
+    {
+      build_date: ENV.fetch('APP_BUILD_DATE'),
```
lostie, line 10: Is there any documentation that describes how/when this variables are set?

dan-palmer: @lostie it's a commonly used across projects and is created as part of the build process. I cant find any documentation though - have requested info from platforms. Will update the commit message when they respond.

dan-palmer:
> {
>   "build_date":"2017-04-11T14:31:18+0000",
>   "build_tag":"pingendpoint.9a6ab35",
>   "commit":"9a6ab3526cbf4b26869be7042398682b6a40b241"
> }

PR #267, opened 2017-04-12
Use action controllers view context to build image tag

dan-palmer:
This fixes the missing placeholder photo when an offender doesn't have a mugshot.

The problem is pre-compilation vs not having pre-compilation of assets in dev & test & the way asset lookup works outside of a controller context. Including the helpers alone that provide the necessary image methods won't work in prod.

This provides similar functionality as that found in the alerts presenter that currently works in production.


lostie:
LGTM 👍

lostie (approved):

PR #270, opened 2017-04-20
Remove current offences through offences model

dan-palmer:
Prior to these changes the offences model acted as a join table -  mapping
collections of current offences & past offences(although they have been
removed recently) to a detainee - this is now unecessary.

We have altered the vocabulary used in the code settling on only
offences - these are the equivalent of what were current offences.
Maintains references to current offences in the UI - we have some copy
changes dropping real soon - changing this might cause some confusion
when those changes get applied.

This work was originally done before the move -> escort refactor took
place. In order to get it work we have decided to maintain the
interfaces that are provided by the risk and healthcare(an previously
the offences models) with regard to the workflow methods. For now we
have wrap the AR offences collection in an object that offers the same
interface as a regular AR collection but also the workflow interfaces
such as the status predicates(not_started?, incomplete?, confirmed?
etc). Ideally this will be removed once more work has been done around
the addition of assessments.

See: #265


db/migrate/20170407140028_migrate_offences_to_current_offences.rb
```diff
@@ -0,0 +1,20 @@
+class MigrateOffencesToCurrentOffences < ActiveRecord::Migration[5.0]
+  def change
+    add_column :current_offences, :detainee_id, :uuid
+    assosciate_existing_current_offences_with_detainee
```
lostie: **[nit pick/not a blocker]** typo in _associate_

lostie (approved):
LGTM 👍

PR #275, opened 2017-04-26
Render healthcare forms from assessment schema

cesidio:
This is the second step into a new approach of managing assessments.

This new approach adds a schema which will be used in the future to describe
all the information for the assessments.

The work in this PR adds the healthcare information into the assessment schema and uses it to render the healthcare forms.

The work in this PR contains also:

- Move shared field partials in app/views/shared
- Add fields_for shared partial to manage multiple medications
- Move select_values_for helper in ApplicationHelper


lostie:
🚫 🙅 Can you please provide a description of what the PR is all about? It's important that other people looking at the PR now or somewhere in the future have a simple way of understanding the background of these changes without having to look through the code.

app/views/shared/_fields_for.html.slim
```diff
@@ -0,0 +1,26 @@
+.multiple
+  div id=question.name
+  div id="error_#{f.object.name}_#{question.name}"
+
+  = f.fields_for question.name do |mf|
+    .multiple-wrapper class=(mf.index == 0 ? 'multiple-first' : 'multiple-not-first')
+      = mf.hidden_field :id
+      - question.subquestions.each do |subquestion|
```
lostie, line 8: 📓 It's worth pointing out that this block is not using the recursive approach of rendering the questions by using it's mapped display_type. The main reason for not doing that right now, is that, the content for this section follows a very specific pattern that follows out of the one defined in the schema.

Might be something we need to re-visit in the future, for the time being it complies with the existent behaviour, which is what we intended to keep.

cesidio: 👍

app/views/shared/_fields_for.html.slim
```diff
@@ -0,0 +1,26 @@
...
+        .multiple-field.inline
+          - if subquestion.answers_options.empty?
+            = mf.text_field subquestion.name.to_sym
+          - else
+            .form-group
+              = mf.label subquestion.name.to_sym, class: 'form-label'
+              = mf.select subquestion.name.to_sym, select_values_for("#{f.object.name}.#{question.name}.#{subquestion.name}"), include_blank: true
```
lostie, line 15: 📓 I think this could be simplified, but there's still quite a lot of dependent models lying around.

Ideally in the future, the presenter for the question, should be able to respond directly to the `values_for` that question.

app/views/shared/_fields_for.html.slim
```diff
@@ -0,0 +1,26 @@
...
+
+      .multiple-field.remove-link.inline
+        .form-group
+          = mf.label :_delete, 'Remove'
+          = mf.check_box :_delete
+
+  .multiple-field.multiple-add
+    = f.submit t("#{f.object.model_class_name}.form.actions.#{question.name}.add_multiple"),
```
lostie, line 23: Same here, this could potentially be simplified, but I can't see atm an obvious way of achieving it without engaging in further refactoring.

lostie (approved):
LGTM 👍

PR #276, opened 2017-04-27
Don't throw an error on nil move date

cesidio:
When we duplicate an escort from an existing one, we copy all
the data from the old one except the move date that we know has
to be different. If we interrupt the process before setting a
date for the move, we end up having escorts with moves with 
a nil date. This fix will prevent the results from the search
on the dashboard from throwing an error.


lostie (approved):
LGTM 👍

PR #277, opened 2017-05-03
Add controlled unlock to risk violence

cesidio:
This PR will add the controlled unlock question to the risk - violence section. There is a also a small refactoring that add a translate_options method for common code that abstracts a common pattern for translating values in multiple choices fields.

[trello ticket](https://trello.com/c/VF9vrj9P/13-as-a-receiver-i-need-to-know-number-of-officers-necessary-to-escort-a-prisoner-in-a-secure-way-control-unlock-so-that-i-keep-me-)


db/migrate/20170428180327_add_controlled_unlock_required_to_risk_to_others.rb
```diff
@@ -0,0 +1,7 @@
+class AddControlledUnlockRequiredToRiskToOthers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :risks, :controlled_unlock_required, :string, default: "unknown"
```
lostie: _Defaults_ in the database for values that are controlled at the _form level_ should not be set.

They're generally a bad practice and show how much we're tying the database schema to the business model data.

lostie (approved):
LGTM 👍

PR #278, opened 2017-05-04
Perform risk form validations using assessment schema

lostie:
[Trello](https://trello.com/c/998LUmye/143-make-changes-to-the-assessments-easier-to-manage)

As another step to centralise the information related with assessments
this PR introduces the following:

- Changes to the new assessment schema to support configurations on how
to validate each of the questions defined in it
- Current risk form implementation (which was 1 form class per risk
section) is replaced by a generic section form class that uses the
information contained in the assessment schemas to perform the
appropriate validations
- Existent behaviour for validations/saving the record was kept on
purpose to make for a gradual migration. What this means in practice is,
for example:
  - that default values 'unknown' were kept
  - before saving the record, any values previously populated that are
no longer required are reset to nil
- Validator classes were added to give support to the current type of
assessment question validations:
  - presence validation
  - inclusion validation
  - date validation
  - group validation


app/form_builders/govuk_elements_errors_helper.rb
```diff
@@ -1,4 +1,42 @@
 module GovukElementsErrorsHelper
+  def self.error_summary(object, heading, description, options = {})
```
lostie, line 2: 📓 I've had a look at the [gem that contains this code](https://github.com/ministryofjustice/govuk_elements_form_builder/blob/4faddb1a9338a17f5e8a0c6e92c94be3834c8b59/app/helpers/govuk_elements_errors_helper.rb) and it is very limited in the way that the display configurations can be defined. Also, it's not in the context of the form (which is really one of the main issues we're trying to solve here), so when we use something like this to define our form:

```ruby
GovukElementsErrorsHelper.error_summary form, 'There were problems saving this form', ''

form_for form, as: 'alias' do |f|
   ...
end
```

The `as: 'alias'` option is not taken into account for displaying the list of errors, making the links to each error point to the wrong anchor :/

Having looked into the source code of the gem, it doesn't seem that a straight simple fix for it, so for the requirements of this PR, I've chosen to keep those changes within our project only.

app/form_builders/mps_form_builder.rb
```diff
@@ -1,4 +1,10 @@
 class MpsFormBuilder < GovukElementsFormBuilder::FormBuilder
+  def error_messages(options = {})
```
lostie, line 2: 📓  This helper allows us to use an inline display of the errors in the context of the form, as such:

```ruby
form_for @object do |f|
  = f.error_messages
end
```

and supports all the additional options that the GovukElementsErrorsHelper supports as well.

app/controllers/risks_controller.rb
```diff
@@ -68,14 +67,7 @@ def redirect_after_update
...
-  end
-
-  def form
-    @_form ||= form_for_risk_section(step).new(risk)
-  end
-
-  def form_for_risk_section(section)
-    "Forms::Risk::#{section.to_s.camelcase}".constantize
+    return unless params[step]
+    params[step].permit!.to_h
```
lostie, line 71: 📓 The `permit!.to_h` was introduced here only as a temporary measure to avoid a [deprecation warning](http://blog.bigbinary.com/2016/07/25/parameters-no-longer-inherit-from-hash-with-indifferent-access-in-rails-5.html) related with the `ActionController::Parameters` and the use of `.with_indifferent_access` on it.

Ideally we should only passed the `permitted` params, but given this PR is introducing a dynamic way of populating the form, supporting the `permitted` params would imply having to add some more methods to the form to returned the `permitted` fields.

That might come in an upcoming PR. 

Also, this is not a major issue at all, as the form by a `reduce` strategy only updates the fields described in the _section schema_.

cesidio (commented):
👍

PR #281, opened 2017-05-08
Use assessments schema to render printout

lostie:
This is yet another step to centralise the management of the assessment
information.

Includes:

- Refactor the PER printout to use the new assessment schema


config/assessments_schema.yml
```diff
@@ -93,93 +93,93 @@ assessment:
                 value: 'no'
               -
                 value: 'unknown'
-      transport:
+      needs:
```
lostie, line 96: 📓 This change is only due to an ordering issue, the needs section has not in the right position.

cesidio (approved):

PR #283, opened 2017-05-08
Update hostage taker text

cesidio:
This PR will also introduce an optional parameter to the validate_as_pdf_that_contains_text matcher to regenerate the PDF file with the actual content.


lostie (approved):
LGTM 👍

PR #284, opened 2017-05-09
Assessment schema driven summary

lostie:
Use assessments schema to render assessment summary page  …
This is yet another step to centralise the assessment management.

This step will replace the existent method to render the assessments
summary page which was using the assessment section specific classes.

Instead it will use the centralised assessment schema to render that
same information.

With this in place the following is now using the assessment schema:

- render the assessment section form page
- validate the answers submitted for an assessment section
- render the assessment printout
- render the assessment summary page


app/presenters/summary/assessment_question_presenter.rb
```diff
@@ -0,0 +1,83 @@
...
+          [
+            detail_label("#{question.name}_collection.#{subquestion.name}"),
+            answer_value(item.send(subquestion.name))
+          ].join(' ')
+        end.join(' | ')
+      end.join('</br>')
+    end
+
+    def detail_label(attribute)
+      I18n.t!(attribute, scope: [:summary, :section, :questions, section_name])
```
cesidio: I think we can get rid of I18n as we are including PresentersHelpers that include ActionView::Helpers::TranslationHelper

cesidio (approved):

PR #286, opened 2017-05-10
Update copy of risk to self section

cesidio:
[trello](https://trello.com/c/5EDhZ0ay/99-1-update-risk-to-self-screen)


lostie (commented):
This is missing a description which at least should container a link to the ticket that it relates to

lostie (approved):
LGTM 👍

PR #289, opened 2017-05-11
Create a new risk of discrimination section

lostie:
According to the new requirements: [Trello](https://trello.com/c/0fES5Z6J/139-3-new-screen-discrimination-against-others)

The existent questions regarding risk of discrimination have been moved out of the violence section into a section of its own.

Also they've been destructured from a nested questions into a set of yes/no, more simple questions.


spec/features/pages/risk.rb
```diff
@@ -34,27 +35,23 @@ def fill_in_risk_from_others
       save_and_continue
     end
 
+    def fill_in_discrimination
+      fill_in_optional_details('Are they a risk to females?', @risk, :risk_to_females)
```
lostie, line 39: 📓 This might be a point for improvement, at the moment I see little advantage having to specify here the exact label if its being managed by the locales.

It seems the DOM would probably require some changes to help us find the elements more accurately without using the current approach.

cesidio (approved):
LGTM

PR #297, opened 2017-05-17
Create an assessment introduction

lostie:
[Trello](https://trello.com/c/0o8Da8TK/165-new-screen-before-you-start)

Using the current assessments schema, this would be most likely the
necessary changes to have an introduction to the assessment that doesn't
really belong to the wizard steps, it's a page shown before that.


cesidio (approved):
very elegant solution, I like it

PR #298, opened 2017-05-17
Spec questionable module without shared example

cesidio:
Given the assessments questions are managed through the schema, we only have a spec for the questionable module itself and test that Healthcare and Risk models include it's behavior


lostie (approved):
LGTM 👍

spec/models/concerns/questionable_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  let(:schema) { Schemas::Assessment.new(hash) }
+
+  subject(:assessment) { TestQuestionableModule.new }
+
+  before { allow(assessment).to receive(:schema).and_return(schema) }
+
+  describe '#total_questions_with_relevant_answer' do
+    it 'returns the number of questions answered with relevance' do
+      assessment.question_1 = 'open'
+      assessment.question_2 = 'yes'
```
lostie: It's not very clear why this is relevant. I know that it is, because I've worked on it, but perhaps this should be a bit more evident in the test file, aka:

```ruby
let(:default_relevant_value) { 'yes' }
```

PR #303, opened 2017-05-22
Change health issues page

cesidio:
These are mostly text changes, the only relevant difference is
the addition of the style to render lists that appear in forms


lostie (commented):
This is missing the link to the related story.

lostie (approved):
LGTM 👍

PR #308, opened 2017-05-23
Add label to the hostage taker date of last incident printout

lostie:
[Trello](https://trello.com/c/IdIb96FG/158-0-5-update-printout-hostage-taker-questions)

**Note:** 

The `I18n.t!(attr)` method should be really `I18n.t(attr, raise: true)`.


cesidio (approved):
LGTM

PR #312, opened 2017-05-25
Wording changes to the healthcare contact

lostie:
[Trello](https://trello.com/c/fHjnoqd6/177-2-update-medical-contact-screen)

- Remove healthcare profession field from healthcare contact section
- Re-worded the remaining fields
- Set a default for the healthcare contact number. Currently this points
to a dummy finder. In the future this should return the matching data
for the prisoner's current establishment.


cesidio (approved):
LGTM

PR #313, opened 2017-05-26
Remove indirection from escort to move models

lostie:
[Trello](https://trello.com/c/bG6jHjqP/47-2-as-someone-adding-new-move-details-to-an-existing-profile-with-a-previous-move-on-moving-people-safely-i-do-not-want-the-appli)

Until now asserting the state of the escort was being delegated to the
move due to the phased migration from the legacy data model.

From now on, escort will respond directly to its current status.

Also, whenever we're re-using an existent escort to create a new one for the same prisoner, the move details are not copied across.


cesidio (approved):
LGTM

PR #314, opened 2017-05-30
Make offences data read only for issued PERs

lostie:
So far, after the PER had been issued it was still possible to attempt
to change the offences information (although, once the form was
submitted an error would be shown).

These changes ensure that the user can no longer attempt to change that
information, and that information is displayed in a table format rather
than in a form as they were until now.

Took the opportunity as well to do some small refactoring in the specs
for the assessment summary pages and extract their common functionality
into a mixin.

### DIFF
--- a/app/views/offences/show.html.slim
+++ b/app/views/offences/show.html.slim
@@ -14,29 +14,39 @@
     header
       h1 Offences
 
-  = GovukElementsErrorsHelper.error_summary form,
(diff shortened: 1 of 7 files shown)


lostie:
This PR ended up being incorporated as part of https://github.com/ministryofjustice/moving-people-safely/pull/315, so closing this one.

PR #320, opened 2017-06-07
Add reviewer info to profile and print page

cesidio:
[Trello](https://trello.com/c/xAaYj9Ip/194-add-date-and-time-stamp-plus-details-of-user-who-confirms-and-saves-the-risk-and-health-sections)

This PR will add the user who confirmed an assessment (only risk and healthcare) and the datetime of when that happened, to the profile page and the print page

### DIFF
--- a/app/views/escorts/_risk.html.slim
+++ b/app/views/escorts/_risk.html.slim
@@ -1,12 +1,12 @@
 #risk
   = render partial: 'status',
-    locals: { workflow: risk,
-              not_started_path: new_escort_risks_path(escort),
-              summary_path: escort_risks_path(escort) }
+    locals: { workflow: risk, not_started_path: new_escort_risks_path(escort), summary_path: escort_risks_path(escort) }
 
   h2
     ' Risk
-    / small 16:44 | 04 Apr 2016
+    - if risk&.reviewed?
+      small
+        = t('assessment.confirmed_timestamp', name: risk.reviewer.full_name, time: risk.reviewed_at.to_s(:humanized))
   - if risk
     .info-table.info-table-big
       .grid-row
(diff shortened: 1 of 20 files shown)


app/views/escorts/_risk.html.slim
```diff
@@ -1,12 +1,12 @@
...
   = render partial: 'status',
-    locals: { workflow: risk,
-              not_started_path: new_escort_risks_path(escort),
-              summary_path: escort_risks_path(escort) }
+    locals: { workflow: risk, not_started_path: new_escort_risks_path(escort), summary_path: escort_risks_path(escort) }
 
   h2
     ' Risk
-    / small 16:44 | 04 Apr 2016
+    - if risk&.reviewer
```
lostie: Where possible we should avoid using this methods in the controllers/views and let the object itself deal with it, having instead something like:

```ruby
risk.reviewed?
```

app/views/escorts/_risk.html.slim
```diff
@@ -1,12 +1,12 @@
...
+      small
+        = t('escort.confirmed_timestamp', name: risk.reviewer.full_name, time: risk.reviewed_at)
   - if risk
```
lostie: Also, the new changes were added outside this condition, but seems more logical to be inside it.

spec/config/date_format_spec.rb
```diff
@@ -9,9 +9,16 @@
   end
 
   describe 'humanized date format' do
-    it 'returns a date in the format DD/MM/YYYY' do
+    it 'returns a date in the format DD MM YYYY' do
```
lostie, line 12: 👍 good 🎯

cesidio: thanks :)

lostie:
LGTM 👍 

A couple of things to keep in mind going forward:

- Description of the commit, specifically on changes like changing the application timezone.
- DRY

PR #321, opened 2017-06-07
Alerts for assessments in review mode

lostie:
[Trello](https://trello.com/c/mPQFLBEW/195-add-reuse-alert-to-risk-summary-and-health-summary)

Adds an alert section to the assessments summary page (risk/healthcare) when the assessment is in  review mode displaying the information of when was the assessment last updated.

This also adds the notion of a **twig/clone** self-relationship on the `Escort` so that we can keep track of which object the escort was cloned from.

**Note to the use of twig as name:**  

From [Wikipedia](https://en.wikipedia.org/wiki/Cloning)

> The term clone, invented by J. B. S. Haldane, is derived from the Ancient Greek word κλών klōn, "twig", referring to the process whereby a new plant can be created from a twig

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,6 @@ group :development, :test do
   gem 'pdftotext'
   gem 'pry-byebug'
   gem 'pry-rails'
-  gem 'timecop'
 end
 
 group :development do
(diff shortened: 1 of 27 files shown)


cesidio (approved):
LGTM 👍

PR #326, opened 2017-06-16
Change Reprint flow to make it harder for users

cesidio:
[Trello](https://trello.com/c/2NfT5Sp9/207-change-reprint-flow-to-make-it-harder-for-users-to-use-this-by-mistake)

This work hides the Reprint link from the dashboard, and when clicking on Reprint on the escort page, the user will be presented with a confirmation screen before he can see the PDF generated.

### DIFF
--- a/app/controllers/escorts/print_controller.rb
+++ b/app/controllers/escorts/print_controller.rb
@@ -1,5 +1,9 @@
 module Escorts
   class PrintController < ApplicationController
+    def new
+      escort
+    end
+
     def show
       error_redirect && return unless printable_escort?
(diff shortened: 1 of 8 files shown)


lostie:
I will state the obvious (which I've already mentioned to the team), that this change is dumb, adding a confirmation page to **re-print** seems to want to fix an issue with how the users interact with the application, but in a bad way.

Also, there's bad patterns we keep on copying across from other pages 🙅 :

- views with business logic
- views using variables with levels of indirection (e.g. a.b.c)

Other than that, LGTM, I guess.

PR #327, opened 2017-06-19
Remove workflows

cesidio:
This refactor adds a reviewable concern and removes the
workflow model.


db/migrate/20170619105503_remove_workflows.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveWorkflows < ActiveRecord::Migration[5.0]
+  def change
+    rename_column :workflows, :workflowable_id, :detainee_id
+    remove_column :workflows, :type
```
lostie: It's a good practice to specify the `type` of the field so that the migrations know how to rollback to the initial state.

lostie:
As discussed offline, I'm still a bit torn regarding this workflow design.

For the time being this DRYs the models that need _reviewing_, so 🚢 it. (we might need to re-visit this workflow design in the future).

PR #328, opened 2017-06-19
Fix date today bug

lostie:
[Trello](https://trello.com/c/7WfCMM2n/211-bug-from-midnight-until-1am-clicking-today-displays-yesterdays-date)

The issue with the date is explained in more detail [here](https://stackoverflow.com/questions/26952421/whats-wrong-with-railss-date).

I've also taken the opportunity to clean up some code related with the way we present the _radio buttons_ for the date picker default options.

### DIFF
--- a/app/models/date_picker.rb
+++ b/app/models/date_picker.rb
@@ -4,12 +4,9 @@ class DatePicker
 
   include Comparable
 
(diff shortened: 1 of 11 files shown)


cesidio (approved):
👍

PR #329, opened 2017-06-20
Move OffencesWorkflow from the Detainee to the Escort

cesidio:
Before this work we had the detainee model having to deal with the offences_workflow data, which is created and modified by the user when the data is not available from the NOMIS API

This work will move the offences_workflow into the escort. 

It will also remove the need to create the offences_workflow before saving the container object (before it was the detainee, now it is the escort).

### DIFF
--- a/db/migrate/20170913110147_offences_workflows_in_escorts.rb
+++ b/db/migrate/20170913110147_offences_workflows_in_escorts.rb
@@ -1,0 +1,15 @@
+class OffencesWorkflowsInEscorts < ActiveRecord::Migration[5.0]
+  def up
+    add_column :offences_workflows, :escort_id, :uuid
+    OffencesWorkflow.all.each { |ow| d = Detainee.find_by_id(ow.detainee_id); ow.update escort_id: d.escort_id if d }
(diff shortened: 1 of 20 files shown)


db/migrate/20170913110147_offences_workflows_in_escorts.rb
```diff
@@ -0,0 +1,8 @@
+class OffencesWorkflowsInEscorts < ActiveRecord::Migration[5.0]
+  def change
+    add_column :offences_workflows, :escort_id, :uuid
+    OffencesWorkflow.all.each { |ow| ow.update escort_id: Detainee.find(ow.detainee_id).escort_id }
+    remove_column :offences_workflows, :detainee_id, :uuid
+    add_index :offences_workflows, :escort_id
+  end
+end
```
doctorpod: These sort of migrations can be dangerous.
1. Try and make them reversible
2. Consider writing output to show progress
3. Consider making this a one off rake task so it does not run as part of deploy - I think we took this approach a few times in PVB

doctorpod (commented):
On the face of it the code looks fine. Not sure what the back story or motivation for this was. Have commented where appropriate.

PR #332, opened 2017-06-22
Improve alerts on Print out

cesidio:
[Trello](https://trello.com/c/4aF1bAvC/49-improve-the-way-alerts-are-understood-by-users-on-print-out)

This work will make the alerts on the print more clear when the value of the field is No/False


app/presenters/print/alert_presenter.rb
```diff
@@ -51,10 +51,14 @@ def toggle_content
     end
 
     def build_toggle_content
-      if toggle.present?
-        h.content_tag(:span, toggle, class: 'alert-toggle')
-      elsif on?
+      return h.content_tag(:span, toggle, class: 'alert-toggle') if toggle.present?
+      if on?
```
lostie: **not a blocker** (more of a tiny improvement), but you could get rid of the if/else completely by turning this if into:

```ruby
on? alert_on_content : alert_off_content
```
and move the content of those to their associated method names.

lostie:
LGTM 👍

PR #333, opened 2017-06-26
Persist issued PERs documents on AWS S3

lostie:
[Trello](https://trello.com/c/DNDPbvch/193-refactor-to-allow-the-addition-deletion-of-questions-without-affecting-previously-issued-pers)

These changes include:

* Adding paperclip gem to manage escort issued documents
* Configuring paperclip for each environment with the appropriate AWS S3
settings where necessary
* Refactor the print controller so that it doesn't generate the PDF for
the issued PER every single time, but instead does the following:
  * Created a service that is solely responsible for issuing the PER,
which includes generating the PDF and storing (either S3 or local
filesystem) and then marking the escort as issued
  * the print controller now calls this service only the escort has not
been issued yet and it is in a valid state to be issued
  * whenever the print action is called for an escort that was already
issued, rather than generating the PDF again, the action just retrieves
the stored PDF and displays it

These changes will improve the reliability of the the issued PER as
the document is no longer mutable (re-generated again)


cesidio (approved):
LGTM

PR #334, opened 2017-06-29
Css improvements print alerts off

cesidio:
[Trello](https://trello.com/c/4aF1bAvC/49-improve-the-way-alerts-are-understood-by-users-on-print-out)

This works improves the way alerts are displayed on the PDF. This will make the label appear in a way that is more similar to what the designer provided us.

Relates to https://github.com/ministryofjustice/moving-people-safely/pull/332


lostie:
LGTM 👍

PR #337, opened 2017-07-03
Allow to cancel PERs

cesidio:
[Trello](https://trello.com/c/MkXUijC8/26-3-as-someone-adding-and-or-completing-a-per-i-want-to-be-able-to-cancel-an-unisssed-move-and-remove-any-associated-data-from-the)

This work will allow users to cancel PERs that have been initiated by my stake

In this work we also delegate methods from escort to associated models in order to comply with Law of Demeter.

### DIFF
--- a/app/controllers/escorts_controller.rb
+++ b/app/controllers/escorts_controller.rb
@@ -1,6 +1,7 @@
 class EscortsController < ApplicationController
   helper_method :escort
   before_action :redirect_if_missing_data, only: :show
+  before_action :redirect_if_not_cancellable, only: [:confirm_cancel, :cancel]
 
   def create
     escort = EscortCreator.call(escort_params)
@@ -16,6 +17,20 @@ class EscortsController < ApplicationController
   def show
   end
 
+  def confirm_cancel
+    @form = Forms::Escort.new(escort)
+  end
+
+  def cancel
+    @form = Forms::Escort.new(escort)
+    if @form.validate(escort_params)
+      escort.cancel!(current_user, escort_params[:cancelling_reason])
+      redirect_to root_path
+    else
+      render :confirm_cancel
+    end
+  end
+
   private
 
   def escort
@@ -23,11 +38,15 @@ class EscortsController < ApplicationController
   end
 
   def escort_params
-    params.require(:escort).permit(:prison_number)
+    params.require(:escort).permit(:prison_number, :cancelling_reason)
   end
 
   def redirect_if_missing_data
     redirect_to(new_escort_detainee_path(escort)) && return unless escort.detainee
     redirect_to(new_escort_move_path(escort)) && return unless escort.move
+  end
+
+  def redirect_if_not_cancellable
+    redirect_to escort_path(escort) unless escort.cancellable?
   end
 end
--- a/app/models/forms/escort.rb
+++ b/app/models/forms/escort.rb
@@ -1,0 +1,7 @@
+module Forms
+  class Escort < Forms::Base
+    property :cancelling_reason, type: String
+
+    validates :cancelling_reason, presence: true
+  end
+end
--- a/app/views/escorts/_actions.html.slim
+++ b/app/views/escorts/_actions.html.slim
@@ -5,5 +5,8 @@ ul.actions
   - elsif escort.completed?
     li
       = link_to_print_in_new_window(escort, styles: 'button button-print')
+  - if escort.cancellable?
+    li
+      = link_to('Cancel PER', confirm_cancel_escort_path(escort), class: 'button')
   li.link-return-home
     = link_to "Return to home", root_path
--- a/app/views/escorts/confirm_cancel.html.slim
+++ b/app/views/escorts/confirm_cancel.html.slim
@@ -1,0 +1,17 @@
+= content_for :breadcrumbs do
+  - breadcrumbs_for_page root: true do
+    - breadcrumb detainee_breadcrumb(escort.detainee), escort_path(escort)
+
+h1
+  = t('.title')
+  br
+  = "#{escort.prison_number}: #{escort.detainee_surname} #{escort.detainee_forenames}"
+
+= form_for @form, url: cancel_escort_path(escort), method: :put do |f|
+  = f.error_messages
+
+  = f.text_area_without_label :cancelling_reason
+
+  .form-actions
+    = f.submit t('.cancel_this_per'), class: 'button'
+    = link_to t('.do_not_cancel_per'), escort_path(escort), class: 'button button-secondary'
--- a/app/views/homepage/_escorts.html.slim
+++ b/app/views/homepage/_escorts.html.slim
@@ -2,25 +2,30 @@
   table
     thead
       tr
-        th.prison_number Prison No.
-        th.name Name
-        th.destination Destination
-        th Risk
-        th Health
-        th Offences
+        th.prison_number= t('.prison_number')
+        th.name= t('.detainee_name')
+        th.destination= t('.destination')
+        th= t('.risk')
+        th= t('.health')
+        th= t('.offences')
         th
     - escorts.each do |escort|
-      tr.move-row id="prison_number_#{escort.detainee.prison_number}"
-        td = link_to escort.detainee.prison_number, escort_path(escort)
+      tr.move-row id="prison_number_#{escort.prison_number}"
+        td = link_to_unless escort.cancelled?, escort.prison_number, escort_path(escort)
         td
           span.strong-text
-            | #{escort.detainee.surname}
+            | #{escort.detainee_surname}
           span
-            |  #{escort.detainee.forenames}
+            |  #{escort.detainee_forenames}
         td = escort.move&.to
         - if escort.issued?
           td.issued colspan=4
-            | Issued
+            = t('.issued')
+        - elsif escort.cancelled?
+          td colspan=4
+            span.cancelled
+              = t('.cancelled')
+            | By: #{escort.canceller_full_name}. #{escort.cancelling_reason}
         - else
           td.completion-status
             span class="#{escort.risk_complete? ? true : false}"
--- a/spec/features/cancelling_a_per_spec.rb
+++ b/spec/features/cancelling_a_per_spec.rb
@@ -1,0 +1,22 @@
+require 'feature_helper'
+
+RSpec.feature 'cancelling a PER', type: :feature do
+  scenario 'Cancelling a non issued PER' do
+    escort = create(:escort, :completed)
+
+    login
+
+    dashboard.assert_escorts_due(1)
+    dashboard.assert_total_detainees_due_to_move_gauge(1)
+
+    visit escort_path(escort)
+
+    escort_page.click_cancel
+
+    fill_in 'escort[cancelling_reason]', with: 'Started by mistake'
+    click_button 'Cancel this PER'
+
+    dashboard.assert_escorts_due(1)
+    dashboard.assert_total_detainees_due_to_move_gauge(0)
+  end
+end
--- a/spec/forms/search_spec.rb
+++ b/spec/forms/search_spec.rb
@@ -44,6 +44,17 @@ RSpec.describe Forms::Search, type: :form do
         end
       end
 
+      context 'when an escort has been cancelled for a given prison number' do
+        let(:prison_number) { 'A1234BC' }
+        let(:detainee) { create(:detainee, prison_number: prison_number) }
+        let!(:escort) { create(:escort, :cancelled, prison_number: prison_number, detainee: detainee) }
+
+        it 'returns nothing' do
+          subject.validate(prison_number: 'A1234BC')
+          expect(subject.detainee).to be_nil
+        end
+      end
+
       context 'when the provided prison name is in a different format than the one recorded for the detainee' do
         let(:prison_number) { 'A1234BC' }
         let(:detainee) { create(:detainee, prison_number: prison_number) }
--- a/spec/requests/escort_cancel_request_spec.rb
+++ b/spec/requests/escort_cancel_request_spec.rb
@@ -1,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe 'PER page requests', type: :request do
+  let(:prison_number) { 'A45345HG' }
+  let(:escort) { create(:escort, prison_number: prison_number) }
+  let(:params) { { escort: { cancelling_reason: 'initiated by mistake' } } }
+
+  describe "#cancel" do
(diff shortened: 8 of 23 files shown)


app/controllers/escorts_controller.rb
```diff
@@ -16,14 +16,28 @@ def create
...
   end
 
+  def confirm_cancel
+    @form = Forms::Escort.new(escort)
+  end
+
+  def cancel
+    @form = Forms::Escort.new(escort)
+    if @form.validate(escort_params)
+      escort.cancel!(current_user, escort_params[:cancelling_reason]) unless escort.issued? || escort.cancelled?
```
lostie: This logic `escort.issued? || escort.cancelled?` should probably be hidden in a method `can_be_cancelled?` so that if the logic changes we don't need to replicate it wherever its being used.

app/controllers/escorts_controller.rb
```diff
@@ -16,14 +16,28 @@ def create
...
+      redirect_to root_path
```
lostie: As discussed offline, when the escort can no longer be cancelled, the user is just redirected to the `root_path` with no information whatsoever of what happened (was the action successful or unsuccessful?). Relying on having a user interface that hides (visually only) the ability to access this action seems bad practice 😰, the action should do whatever needs to do regardless of the user interface associated with it.

app/models/forms/escort.rb
```diff
@@ -0,0 +1,7 @@
+module Forms
+  class Escort < Forms::Base
```
lostie, line 2: This should probably be renamed to match what the form is doing, e.g `CancelEscort`. Seems to be a very specific action to be put into the most generic `Escort` form that normally collects a set of attributes for the entire escort.

lostie, line 2: Is there any specific need to inherit from `Forms::Base` ? If not, I would suggest to just have the form class to use the `ActiveModel::Validations` mixin instead.

app/views/homepage/_escorts.html.slim
```diff
@@ -2,25 +2,30 @@
...
     thead
       tr
-        th.prison_number Prison No.
-        th.name Name
-        th.destination Destination
-        th Risk
-        th Health
-        th Offences
+        th.prison_number= t('.prison_no')
+        th.name= t('.name')
```
lostie: Who's name is this? the escort name?

app/views/homepage/_escorts.html.slim
```diff
@@ -2,25 +2,30 @@
   table
     thead
       tr
-        th.prison_number Prison No.
-        th.name Name
-        th.destination Destination
-        th Risk
-        th Health
-        th Offences
+        th.prison_number= t('.prison_no')
```
lostie: Any reason for having a shortcut(ed) name for a locale? Saving money on characters?

app/views/escorts/confirm_cancel.html.slim
```diff
@@ -0,0 +1,18 @@
+= content_for :breadcrumbs do
+  - breadcrumbs_for_page root: true do
+    - breadcrumb detainee_breadcrumb(escort.detainee), escort_path(escort)
+
+.escort.per-page
+  h1
+    = t('.cancelling_per_for')
```
lostie: Isn't this just the title for the cancel escort form? Is there a need to have the locale key to match what the actual translation is? Shouldn't this just be something like `form.cancel_escort.title` or something similar?

app/views/escorts/confirm_cancel.html.slim
```diff
@@ -0,0 +1,18 @@
+= content_for :breadcrumbs do
+  - breadcrumbs_for_page root: true do
+    - breadcrumb detainee_breadcrumb(escort.detainee), escort_path(escort)
+
+.escort.per-page
```
lostie: Is this CSS needed for this page? This CSS was specifically introduced to the PER summary page.

app/views/escorts/_actions.html.slim
```diff
@@ -5,5 +5,8 @@ ul.actions
   - elsif escort.completed?
     li
       = link_to_print_in_new_window(escort, styles: 'button button-print')
+  - unless escort.issued?
```
lostie: Just above this code we have the reverse query `- if escort.issued?`. Can't this be handled by a single condition?

db/migrate/20170629152622_add_deleter_to_escort.rb
```diff
@@ -0,0 +1,7 @@
+class AddDeleterToEscort < ActiveRecord::Migration[5.0]
```
lostie: Shouldn't the name of the migration be something like "AddCancellerToEscort" or "AddCancellationColumnsToEscorts" ?

lostie: If you're using one this columns to scope the results of a query, should you consider to have an index on it?

spec/features/cancelling_a_per_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    escort = create(:escort, :completed)
+
+    visit escort_path(escort)
+
+    escort_page.click_cancel
+
+    fill_in 'escort[cancelling_reason]', with: 'Started by mistake'
+    click_button 'Cancel this PER'
+
+    dashboard.assert_escorts_due(1)
```
lostie, line 19: Should this also be asserted before the cancellation so that we know something actually changed?

spec/features/cancelling_a_per_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'feature_helper'
+
+RSpec.feature 'cancelling a PER', type: :feature do
+  scenario 'Cancelling a non issued PER' do
```
lostie, line 4: This covers only the happy scenario. Should we have a `request` spec for the cancelling action?

cesidio: I added a request spec that covers the unhappy scenario

app/controllers/escorts_controller.rb
```diff
@@ -1,6 +1,7 @@
 class EscortsController < ApplicationController
   helper_method :escort
   before_action :redirect_if_missing_data, only: :show
+  before_action :redirect_if_issued_or_cancelled_escort, only: [:confirm_cancel, :cancel]
```
lostie: If tomorrow there's a third state that also matches this criteria, this method name will need to be changed. The name of the method shouldn't depend on the internal logic of it.

Consider instead something like `:redirect_if_not_cancellable`.

app/controllers/escorts_controller.rb
```diff
@@ -16,18 +17,36 @@ def create
...
+    params.require(:escort).permit(:prison_number, :cancelling_reason)
   end
 
   def redirect_if_missing_data
     redirect_to(new_escort_detainee_path(escort)) && return unless escort.detainee
     redirect_to(new_escort_move_path(escort)) && return unless escort.move
   end
+
+  def redirect_if_issued_or_cancelled_escort
+    redirect_to escort_path(escort) if escort.issued? || escort.cancelled?
```
lostie: 2 things:

- Can we add a error message to the redirect which indicates to the user what's happening?
- The logic that decides if the escort is cancellable should be probably hidden in a method like `is_cancellable?` and have this code just have:

  `redirect_to escort_path(escort) unless escort.is_cancellable?`

app/views/escorts/_actions.html.slim
```diff
@@ -5,5 +5,8 @@ ul.actions
   - elsif escort.completed?
     li
       = link_to_print_in_new_window(escort, styles: 'button button-print')
+  - unless escort.issued? || escort.cancelled?
```
lostie: 🙅 This is exactly the reason why we shouldn't keep on repeating logic all over the application that should be in the escort. This logic already exists in the escort controller. Can we please follow this [comment](https://github.com/ministryofjustice/moving-people-safely/pull/337/files#r125505777) suggestion pls to DRY this?

app/views/escorts/confirm_cancel.html.slim
```diff
@@ -0,0 +1,17 @@
+= content_for :breadcrumbs do
+  - breadcrumbs_for_page root: true do
+    - breadcrumb detainee_breadcrumb(escort.detainee), escort_path(escort)
+
+h1
+  = t('.cancelling_per_for')
```
lostie: This [comment](https://github.com/ministryofjustice/moving-people-safely/pull/337#discussion_r125449193) I think is still relevant here. Why are we creating random locale names dependant on the actual locale translation? 

IMO, this is the most basic stuff we do on locales: they have a context, the context should be pretty clear when I'm looking at it in the locales files. This one, on the other hand is anything but clear. It's clear for me it's the title of the form, I'm just missing to understand why we don't keep the locales to this very clear namespaces...

spec/forms/search_spec.rb
```diff
@@ -44,6 +44,17 @@
         end
       end
 
+      context 'when an escort has been cancelled for a given prison number' do
+        let(:prison_number) { 'A1234BC' }
+        let(:detainee) { create(:detainee, prison_number: prison_number) }
+        let!(:escort) { create(:escort, prison_number: prison_number, detainee: detainee, cancelled_at: Time.now) }
```
lostie: Can we please avoid using time values that are not in UTC within tests? We had issues before with this because the CI environment might be configured in a different timezone.

Also I would recommend this to be moved into a **trait** within the _escort_ factory so that can be re-used and it's an escort in a valid cancelled state (atm, has no associated user or reason)

spec/requests/escort_cancel_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      context 'when the escort for the provided id does not exist' do
+        it 'raises a record not found error' do
+          expect {
+            put "/escorts/#{SecureRandom.uuid}/cancel", params: params
+          }.to raise_error(ActiveRecord::RecordNotFound)
+        end
+      end
+
+      context "when the escort has already been cancelled" do
+        let(:escort) { create(:escort, prison_number: prison_number, cancelled_at: Time.now) }
```
lostie: Please use a trait as mentioned [here](https://github.com/ministryofjustice/moving-people-safely/pull/337/files#r125508392). DRY

spec/requests/escort_cancel_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        it 'redirects to the escort page' do
+          put "/escorts/#{escort.id}/cancel", params: params
+          expect(response).to have_http_status(302)
+          expect(response).to redirect_to escort_path(escort)
+        end
+      end
+
+      context "when the escort has already been issued" do
+        let(:escort) { create(:escort, prison_number: prison_number, issued_at: Time.now) }
```
lostie: [Ditto](https://github.com/ministryofjustice/moving-people-safely/pull/337/files#r125508692)

spec/requests/escort_cancel_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        it 'redirects to the escort page' do
+          put "/escorts/#{escort.id}/cancel", params: params
+          expect(response).to have_http_status(302)
+          expect(response).to redirect_to escort_path(escort)
+        end
+      end
+
+      context "with a valid escort ID" do
+        it "responds with 200" do
```
lostie: 🙋 The description of the test doesn't match the expectation. Something is wrong here, either one of the other!

spec/services/create_escort_spec.rb
```diff
@@ -14,18 +14,32 @@
...
+        expect_risk_assessment_to_be_cloned(existent_escort, escort)
+        expect_healthcare_assessment_to_be_cloned(existent_escort, escort)
+        expect_offences_to_be_cloned(existent_escort, escort)
+        expect(escort.move).to be_nil
+        expect(escort.twig).to eq(existent_escort)
+      end
+    end
+
+    context 'and escort is cancelled' do
+      let!(:existent_escort) { create(:escort, :completed, prison_number: prison_number, cancelled_at: Time.now) }
```
lostie: Please use a trait for `:cancelled`. I've seen the same code repeated now in a lots of specs. DRY.

spec/services/create_escort_spec.rb
```diff
@@ -14,18 +14,32 @@
...
+        expect_detainee_to_be_cloned(existent_escort, escort)
+        expect_risk_assessment_to_be_cloned(existent_escort, escort)
+        expect_healthcare_assessment_to_be_cloned(existent_escort, escort)
+        expect_offences_to_be_cloned(existent_escort, escort)
+        expect(escort.move).to be_nil
+        expect(escort.twig).to eq(existent_escort)
+      end
+    end
+
+    context 'and escort is cancelled' do
```
lostie, line 33: ~How is there a change in behaviour for this service if the service itself and nothing inside it seems to have been affected/changed?~

I've seen where the change is, sorry 😰

spec/services/create_escort_spec.rb
```diff
@@ -14,18 +14,32 @@
...
+      let!(:existent_escort) { create(:escort, :completed, prison_number: prison_number, cancelled_at: Time.now) }
+
+      it 'create a new escort' do
```
lostie: This scenario is the same as the one for when there's no escorts at all for the given prison number.

Also this actual only covers part of the scenario, as if there's at least two escorts for that prison number, one cancelled, the other one issued, this service should clone the issued one. That scenario is missing in this spec.

app/views/escorts/confirm_cancel.html.slim
```diff
@@ -0,0 +1,17 @@
+= content_for :breadcrumbs do
```
lostie, line 1: Isn't a bit odd that we now have two confirmation pages (re-printing - 🤣 🏃 I still can take this a real feature -  and cancelling) and they are managed through the controller/views in a completely different way? Seems like instead of bringing consistency to the app, we're doing the opposite.

spec/features/cancelling_a_per_spec.rb
cesidio: Added it

app/views/escorts/confirm_cancel.html.slim
cesidio: I see your point. The thing is that they are both confirmation pages, but they work differently. In the case of cancelling a PER, you have to provide a reason. In the case of reprint, we just want to make reprint more difficult for users.

app/models/forms/escort.rb
cesidio: The reason is because we are using the MpsFormBuilder for the text_area_without_label method as it comes handy in this case.

PR #352, opened 2017-07-20
Add font install instructions to README

doctorpod:


.gitignore
```diff
@@ -31,3 +31,6 @@ coverage
 
 # Ignore ruby gemset files
 .ruby-gemset
+
+# Ignore OSX specific artifacts
```
lostie: These kind of ignore files should normally be in our personal [git global ignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore), they shouldn't be necessary to be added here.

lostie:
LGTM 👍

.gitignore
doctorpod: Ah nice one - will revert this bit

PR #376, opened 2017-09-11
Add logic to forbid editing a cancelled escort

cesidio:


app/views/offences/show.html.slim
```diff
@@ -14,7 +14,7 @@
     header
       h1 Offences
 
-  - if escort.issued?
+  - if !escort.editable?
```
doctorpod: Use `unless` so we can get rid of the `!`

doctorpod (approved):

PR #382, opened 2017-09-27
Remove support for subsections

cesidio:


app/models/forms/assessments/section.rb
```diff
@@ -39,8 +32,13 @@ def scope
...
       def valid?
-        [valid_subsections?, valid_answers?].inject(true) do |valid, valid_component|
-          valid_component && valid
+        answers.inject(true) do |valid, answer|
+          (answer.valid? && valid).tap do
+            answer.errors.each do |attr, message|
+              errors.add(attr, message)
+            end
+            answer.errors.clear
+          end
```
doctorpod, line 41: `inject` always gives me a total brain fart (but keep using it).
Maybe rename *valid* to something like *ok_so_far*. 
It took ma a while to understand what you were doing with `tap` here. I see what you're doing now: you want the boolean expression as the last line of the block - not using `tap` would result in more lines I suppose... Carry on.

doctorpod (commented):
Nice refactor - lots of code removed :)

PR #385, opened 2017-10-12
Defined locations

doctorpod:
Ticket: https://trello.com/c/NTRXMgJr


spec/support/fixtures/pdf-text-all-no-answers.txt
```diff
@@ -27,7 +27,7 @@ British           Baptist            European
 
 Move details
 Date of travel                       From                                  To
-22 Apr 2099                          HMP Bedford                           Luton Crown Court
+22 Apr 2099                                                                Luton Crown Court
```
cesidio: Why don't we have the From field filled in?

spec/support/fixtures/pdf-text-all-yes-answers.txt
```diff
@@ -27,7 +27,7 @@ British           Baptist            European
 
 Move details
 Date of travel                       From                                  To
-22 Apr 2099                          HMP Bedford                           Luton Crown Court
+22 Apr 2099                                                                Luton Crown Court
```
cesidio: same here

app/views/moves/_form.html.slim
```diff
@@ -8,7 +8,31 @@
...
       = from_establishment&.name
-    = f.text_field :to
+
+    .radio-concertina
+      .form-group
+        fieldset
+          legend
+            span.form-label-bold To
+
+          - ::Establishment::ESTABLISHMENT_TYPES.each do |option|
```
cesidio: Does it need the starting :: ?
::Establishment::ESTABLISHMENT_TYPES -> Establishment::ESTABLISHMENT_TYPES

app/models/forms/move.rb
```diff
@@ -41,5 +41,34 @@ def not_for_release_reasons
     def not_for_release_reason_with_details
       REASON_WITH_DETAILS
     end
+
+    FREE_FORM_DESTINATION_TYPES = %i[hospital other].freeze
+
+    property :to_type, validates: { presence: true }
+
+    ::Establishment::ESTABLISHMENT_TYPES.each do |establishment_type|
```
cesidio, line 49: ::Establishment::ESTABLISHMENT_TYPES -> Establishment::ESTABLISHMENT_TYPES

app/views/moves/_form.html.slim
```diff
@@ -8,7 +8,31 @@
...
+            = f.radio_concertina_option(:to_type, I18n.t(option, scope: 'move.destination_types.labels'), option) do
```
cesidio: In the view you can omit I18n:
I18n.t -> t

app/views/moves/_form.html.slim
```diff
@@ -8,7 +8,31 @@
...
+              label.form-label for=option
+                = I18n.t(option, scope: 'move.destination_types.prompts')
```
cesidio: I18n.t -> t

app/views/moves/_form.html.slim
```diff
@@ -8,7 +8,31 @@
...
+                  .span.error-message
+                    = form.errors["to_#{option}"].join(', ')
+                  = f.select "to_#{option}", option.to_s.classify.constantize.pluck(:name), {include_blank: true}, { id: option }
+              - else
+                = f.select "to_#{option}", option.to_s.classify.constantize.pluck(:name), {include_blank: true}, { id: option }
+
+          - Forms::Move::FREE_FORM_DESTINATION_TYPES.each do |option|
+            = f.radio_concertina_option(:to_type, I18n.t(option, scope: 'move.destination_types.labels'), option) do
+              label.form-label for="#{option}-text"
+                = I18n.t(option, scope: 'move.destination_types.prompts')
```
cesidio: I18n.t -> t

db/migrate/20171010154443_add_to_type_to_moves.rb
```diff
@@ -0,0 +1,36 @@
+class AddToTypeToMoves < ActiveRecord::Migration[5.0]
+  def up
+    add_column :moves, :to_type, :string
+
+    say_with_time "Evaluating `to_type` in moves..." do
+      moves = Move.all
+
+      moves.each do |move| 
+        unless move.to.nil?
```
cesidio: Not that important as this is run only once, but
items.each do |item|
  unless condition -> next if
    do something
  end
end

cesidio: so you do not indent one more layer and the code is more legible

app/views/moves/_form.html.slim
doctorpod: Yes - it failed to load without it

app/models/forms/move.rb
doctorpod: Without the leading `::` we get `NameError: wrong constant name move` when we run the specs

cesidio: 👍

app/views/moves/_form.html.slim
```diff
@@ -8,7 +8,31 @@
...
+
+          - Establishment::ESTABLISHMENT_TYPES.each do |option|
+            = f.radio_concertina_option(:to_type, t(option, scope: 'move.destination_types.labels'), option) do
+              label.form-label for=option
+                = t(option, scope: 'move.destination_types.prompts')
+              - if form.errors["to_#{option}"].any?
+                .error
+                  .span.error-message
+                    = form.errors["to_#{option}"].join(', ')
+                  = f.select "to_#{option}", option.to_s.classify.constantize.pluck(:name), {include_blank: true}, { id: option }
```
cesidio, line 26: option.to_s.classify.constantize.pluck(:name, :id)

with this we can still show name in the autocomplete, but store ids in the database. I would prefer this approach to store ids over strings

cesidio: After discussing on it I am happy with the current approach

cesidio (approved):

PR #395, opened 2017-11-20
Show 'no known risks' when risks section empty

doctorpod:
Ticket: https://trello.com/c/dNDPcXhV

### DIFF
--- a/app/models/concerns/questionable.rb
+++ b/app/models/concerns/questionable.rb
@@ -91,7 +91,7 @@ module Questionable
     end
 
     def any_questions_answered?
-      sections.flat_map(&:questions).flat_map(&:relevant_answer?).any?
+      mandatory_questions.any?(&:answered?)
     end
   end
 end
--- a/app/models/concerns/questionable.rb
+++ b/app/models/concerns/questionable.rb
@@ -89,5 +89,9 @@ module Questionable
     def all_questions_answered?
       mandatory_questions.all?(&:answered?)
     end
+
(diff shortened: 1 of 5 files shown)


app/models/concerns/questionable.rb
```diff
@@ -89,5 +89,9 @@ def mandatory_questions
     def all_questions_answered?
       mandatory_questions.all?(&:answered?)
     end
+
+    def any_questions_answered?
```
cesidio: What about?:

```
   def no_questions_answered?		
      mandatory_questions.none?(&:answered?)		
    end
```

doctorpod: Have refactored the body of the method to more closely match your example but kept the name the same. Calling `any_questions_answered?` from the view is more readable than `!no_questions_answered?`.

cesidio (approved):
👍

app/views/escorts/_assessment.html.slim
```diff
@@ -12,7 +12,7 @@ section id=name.downcase
       p.review
         = t('assessment.confirmed_timestamp', name: assessment.reviewer.full_name, time: assessment.reviewed_at.to_s(:humanized))
 
-  - if assessment
+  - if assessment && assessment.any_questions_relevant?
```
cesidio: usually rubocop complains about not using & operator
assessment&.any_questions_relevant?

cesidio (approved):

PR #407, opened 2017-12-07
Exclude prisons with old end dates

doctorpod:
Ticket: https://trello.com/c/IdZWhX5J

### DIFF
--- a/db/seeds/populate_prisons.rb
+++ b/db/seeds/populate_prisons.rb
@@ -13,7 +13,6 @@ module Seeds
 
     def call
       logger.info "Seeding #{prison_seeds.size} prisons..."
-      Prison.unscoped.delete_all
       seeded = 0
       prison_seeds.each do |prison_seed|
         prison = seed_entry(prison_seed)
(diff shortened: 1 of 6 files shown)


cesidio (approved):

db/seeds/populate_prisons.rb
```diff
@@ -27,9 +28,9 @@ def call
 
     def seed_entry(prison_seed)
       entry = Entries::Prison.new(prison_seed)
-      Prison.create_with(entry.to_h).find_or_create_by(nomis_id: entry.nomis_id)
-    rescue
-      logger.error "Error seeding #{prison_seed.inspect}"
+      Prison.create(entry.to_h)
```
cesidio: This change will force the removal and recreation of all the establishments, leaving a broken external ID in the moves:

from_establishment_id

I like the approach of changing the seed, but in order to do that we would need all the logic to change from establishment_id to nomis_id in the moves tables.

cesidio (changes requested):

cesidio (approved):

PR #417, opened 2018-01-02
Consistent alerts

cesidio:


app/models/risk.rb
```diff
@@ -7,22 +7,31 @@ class Risk < ApplicationRecord
...
+  def alerts
+    alerts = {}
+    alerts[:acct_status] = acct_status_alert_on?
+    alerts[:self_harm] = (self_harm == 'yes' || acct_status_alert_on?)
+    alerts[:rule_45] = (rule_45 == 'yes')
+    alerts[:current_e_risk] = (current_e_risk == 'yes' || previous_escape_attempts == 'yes')
+    alerts[:csra] = (csra == 'high')
+    alerts[:category_a] = (category_a == 'yes')
+    alerts
   end
```
doctorpod: I actually prefer a literal hash approach:
```ruby
def alerts
  {
    acct_status: acct_status_alert_on?,
    self_harm: (self_harm == 'yes' || acct_status_alert_on?),
    ...
  }
end
```
I think it's slightly more readable.

doctorpod (approved):

PR #423, opened 2018-01-08
Reload offences at escort clone

doctorpod:
Ticket: https://trello.com/c/TEYHiz2m


app/services/escort_creator.rb
```diff
@@ -45,4 +50,21 @@ def deep_clone_escort
       except: EXCEPT_GRAPH
     )
   end
+
+  def refresh_offences(escort)
+    if nomis_offences.any? && escort.offences.present?
```
cesidio: The main problem in moving the logic to retrieve offences at the time of escort creation is that we do not have a detainee object created at this stage if we are creating a PER for the first time, so escort.offences.present? will always be nil and we do not populate offences ever. Before, when this logic was in the offences controller, we had a detainee object created by that time. We have to think for a different strategy.

cesidio (changes requested):

doctorpod:
Cesi (and I) came up with a simpler approach (https://github.com/ministryofjustice/moving-people-safely/pull/431)

PR #424, opened 2018-01-15
Make hostage taker last incident dates optional

doctorpod:
Ticket: https://trello.com/c/QmhEgmkv

### DIFF
--- a/db/migrate/20180111084643_change_hostage_taker_dates_to_strings.rb
+++ b/db/migrate/20180111084643_change_hostage_taker_dates_to_strings.rb
@@ -1,0 +1,87 @@
+class ChangeHostageTakerDatesToStrings < ActiveRecord::Migration[5.0]
+  def up
+    COLUMNS.each do |column|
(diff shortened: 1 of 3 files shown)


db/migrate/20180111084643_change_hostage_taker_dates_to_strings.rb
```diff
@@ -0,0 +1,91 @@
...
+          "#{append_text_to_name(column)}=".to_sym,
+          risk.send(column).to_s
+        )
+      end
+    end
+
+    risk.save
+  end
+
+  def update_date_columns(risk)
```
cesidio: In https://github.com/ministryofjustice/moving-people-safely/pull/388/files
we used a different approach, we called update_column method to not change the timestamps of the models. This approach will change the timestamps. It is not a big deal, but we lose when the last real update happened.

cesidio (changes requested):

cesidio (approved):

cesidio (approved):

PR #450, opened 2018-03-19
Refactor using existing risk fetcher and mapper

doctorpod:


db/schema.rb
```diff
@@ -167,11 +167,8 @@
     t.text "intimidation_to_witnesses_details"
     t.string "hostage_taker"
     t.boolean "staff_hostage_taker"
-    t.date "date_most_recent_staff_hostage_taker_incident"
```
cesidio: This needs to be removed from this PR as it is not relevant and will cause conflicts later on.

cesidio (changes requested):

cesidio (approved):

PR #462, opened 2018-04-18
Hide cancel PER button from court users

cesidio:


app/views/escorts/_actions.html.slim
```diff
@@ -5,7 +5,7 @@ ul.actions
   - elsif escort.issued?
     li
       = link_to('Reprint', new_escort_print_path(escort), class: 'button button-print')
-  - if !escort.cancelled? && !escort.expired?
+  - if !escort.cancelled? && !escort.expired? && !current_user.court?
```
doctorpod, line 8: Is it the case that court users cannot cancel a PER at all? If so would there be controller restrictions?

doctorpod (approved):

PR #467, opened 2018-05-08
Remove Assessment schema and use a more Rails way

cesidio:


db/schema.rb
```diff
@@ -112,6 +121,12 @@
     t.index ["escort_id"], name: "index_moves_on_escort_id"
   end
 
+  create_table "must_not_return_details", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
```
doctorpod, line 124: What does this do?

doctorpod (commented):
My only suggestion would be to consider renaming the `views/healthcare/sections/escort` folder to `screen` to more closely match it's `print` and `summary` siblings -- but entirely optional

db/schema.rb
cesidio: We were handling nested objects (medications, must not return details) in a serialised field within healthcare or risk. With this we go back to normal relational database tables.

PR #472, opened 2018-05-22
Add violence section

doctorpod:
Ticket: https://trello.com/c/dDVbYZox

### DIFF
--- a/app/models/risk.rb
+++ b/app/models/risk.rb
@@ -2,11 +2,11 @@ class Risk < ApplicationRecord
   include Questionable
   include Reviewable
 
-  SECTIONS = %w[risk_to_self segregation security harassment_and_gangs discrimination escape
+  SECTIONS = %w[risk_to_self segregation security violent_or_dangerous harassment_and_gangs discrimination escape
                 hostage_taker sex_offences concealed_weapons arson return_instructions other_risk].freeze
   MANDATORY_QUESTIONS = %w[acct_status self_harm csra rule_45 vulnerable_prisoner controlled_unlock
                            category_a high_profile pnc_warnings intimidation_public intimidation_prisoners
-                           gang_member violence_to_staff risk_to_females homophobic racist
+                           violent_or_dangerous gang_member violence_to_staff risk_to_females homophobic racist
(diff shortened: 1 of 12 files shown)


spec/support/fixtures/pdf-text-all-yes-answers.txt
```diff
@@ -62,6 +62,7 @@ Security                             Yes
   Cat A or potential Cat A           Yes
   High public interest               Yes                 High profile details
   PNC warnings                       Yes                 Pnc warning details
+Violence                             No
```
cesidio: This is inconsistent with the rest of the spec. As the name of the spec suggests, we set all the values to Yes, and violence section should follow the same pattern

spec/features/printing_spec.rb
```diff
@@ -161,7 +161,9 @@
           other_risk: 'yes',
           other_risk_details: 'suspected terrorist',
           reviewer: reviewer,
-          reviewed_at: DateTime.civil(2016, 3, 10, 12, 30)
+          reviewed_at: DateTime.civil(2016, 3, 10, 12, 30),
+          violent_or_dangerous: 'yes',
```
cesidio: I would suggest to keep the order of the fields, so that it is easier to read.

app/models/risk.rb
```diff
@@ -2,7 +2,7 @@ class Risk < ApplicationRecord
   include Questionable
   include Reviewable
 
-  SECTIONS = %w[risk_to_self segregation security harassment_and_gangs discrimination escape
+  SECTIONS = %w[risk_to_self segregation security violent_or_dangerous harassment_and_gangs discrimination escape
                 hostage_taker sex_offences concealed_weapons arson return_instructions other_risk].freeze
   MANDATORY_QUESTIONS = %w[acct_status self_harm csra rule_45 vulnerable_prisoner controlled_unlock
```
cesidio, line 7: You need to add violent_or_dangerous to the MANDATORY_QUESTIONS, otherwise the assessment can responds to all_questions_answered? even when this question is not answered.

cesidio (changes requested):

cesidio (approved):

PR #473, opened 2018-05-22
Remove presenters layer and use vanilla rails helpers

cesidio:


app/helpers/escorts_helper.rb
```diff
@@ -0,0 +1,101 @@
...
+  def date_of_birth(detainee)
+    detainee.date_of_birth.to_s(:humanized)
+  end
+
+  def gender_code(detainee)
+    detainee.gender == 'male' ? 'M' : 'F'
+  end
+
+  def short_ethnicity(detainee)
+    return 'White: British' if detainee.ethnicity == 'White: Eng./Welsh/Scot./N.Irish/British'
```
doctorpod, line 24: Ideally we should always use I18n for these snippets - maybe a future ticket

app/views/escorts/print/cover.html.slim
```diff
@@ -28,83 +28,83 @@ html
...
-            span#detainee-identifier.strong-text => detainee.identifier
-            = detainee.forenames
+            span#detainee-identifier.strong-text => identifier(escort.detainee)
+            = escort.detainee.forenames
         div.grid-row.indicators
           #not_for_release_alert.grid-column.column-one-half
-            = render partial: 'escorts/print/alert', locals: { alert_name: :not_for_release, on: alerts[:not_for_release] }
-            - if alerts[:not_for_release]
-              p.alert-text= move.not_for_release_text
+            = render partial: 'escorts/print/alert', locals: { alert_name: :not_for_release, on: escort.alerts[:not_for_release] }
```
doctorpod, line 35: All these `render partial: 'foo', locals: {bar: 'bish', baz: 'bosh'}` calls could be shortened to `render 'foo', bar: 'bish', baz: 'bosh'`

doctorpod (approved):
The app has lost a lot of weight - nice - just a couple of optional suggestions - otherwise all good

app/views/escorts/print/cover.html.slim
cesidio: 👍

PR #478, opened 2018-05-30
Make offences belong to escort instead of detainee

cesidio:


app/models/offence.rb
```diff
@@ -1,3 +1,3 @@
 class Offence < ApplicationRecord
-  belongs_to :detainee
+  belongs_to :escort
```
doctorpod, line 2: Don't we need a migration to populate this?

doctorpod (changes requested):

app/models/offence.rb
cesidio: There is a migration

doctorpod (approved):

PR #481, opened 2018-06-05
Allow search from police

cesidio:


app/models/forms/search.rb
```diff
@@ -4,12 +4,31 @@ class Search
...
+      @prison_number = search[:prison_number]&.upcase
+      @pnc_number = search[:pnc_number]&.upcase
+    end
+
+    def identifier
+      prison_number || pnc_number
+    end
+
+    def identifier_type
+      return 'PNC number' if pnc_number
```
doctorpod, line 29: It's not only here but in several places, we have bare strings that really should be via I18n - maybe fix later

app/services/escort_creator.rb
```diff
@@ -43,7 +44,7 @@ def call
   ].freeze
 
   def existent_escort
-    @existent_escort ||= Escort.uncancelled.find_by(prison_number: prison_number)
+    @existent_escort ||= Escort.uncancelled.find_by(prison_number: prison_number) if prison_number.present?
```
doctorpod: What if `pnc_number` is present?

app/views/search/_result.html.slim
```diff
@@ -23,7 +23,10 @@ table
 .form-wrapper
   - if @escort.issued?
     = form_for Escort.new(prison_number: @escort.prison_number), url: escorts_path, method: :post do |f|
-      = f.hidden_field :prison_number
+      - if current_user.police?
+        = f.hidden_field :prison_number
```
doctorpod: Should this be `pnc_number`?

spec/factories/escort_factory.rb
```diff
@@ -6,12 +6,17 @@
       [a[0],b[0],b[1],b[2],b[3],a[1],a[2]].join
     end
 
+    pnc_number do
+      a = 1.times.map { ('A'..'Z').to_a.sample }
+      b = 7.times.map { (0..9).to_a.sample }
+      [b[0],b[1],"/",b[2],b[3],b[4],b[5],a[0]].join
```
doctorpod: Or
```
[
  '02d' % rand(99),
  '/',
  '04d' % rand(9999),
  ('A'..'Z').to_a.sample
].join
```

app/models/forms/search.rb
cesidio: I agree on this, we should move all these into translations

PR #483, opened 2018-06-07
Ensure admins can see all PERs

doctorpod:
Ticket: https://trello.com/c/ZAxQujJC


app/models/user.rb
```diff
@@ -59,6 +54,12 @@ def can_access_escort?(escort)
     authorized_establishments.any? { |establishment| establishment.sso_id == escort_establishment_sso_id }
   end
 
+  ADMIN_ORGANISATION = 'digital.noms.moj'.freeze
```
cesidio, line 57: I like constants defined at the beginning of the file, but that's just personal preference. I can see why you moved them here.

cesidio (approved):

PR #493, opened 2018-06-21
Change healthcare risks and offences sections for police

cesidio:


app/models/risk.rb
```diff
@@ -31,7 +21,7 @@ def alerts
   end
 
   def relevant_questions
-    @relevant_questions ||= MANDATORY_QUESTIONS.select do |question|
+    @relevant_questions ||= Risk.mandatory_questions(location).select do |question|
```
doctorpod: Isn't the `Risk.` redundant?

doctorpod (approved):

app/form_builders/mps_form_builder.rb
```diff
@@ -122,21 +122,28 @@ def fieldset_legend(attribute, options = {})
       if options.fetch(:legend, true)
         tags << content_tag(
           :span,
-          fieldset_text(attribute),
+          location_text(fieldset_text(attribute)),
```
doctorpod, line 135: Ah! Is this the bit that replaces my 46,000 lines of code?

spec/features/filling_in_a_police_per_spec.rb
```diff
@@ -13,10 +13,11 @@
 
     expect(current_path).to eq root_path
 
-    healthcare_data = build(:healthcare, :with_medications)
-    risk_data = build(:risk, :with_high_csra)
-    detainee = build(:detainee)
     move_data = build(:move, from_establishment: banbury_police_station)
+    escort = build(:escort, move: move_data)
```
doctorpod, line 26: I wonder if we should consider at some staging using the rspec-examples gem to make these feature specs more readable? (https://github.com/railsware/rspec-example_steps). I do like the Given, When, Then syntax. We have a lot of boilerplate setup which could be in a Given block for example.

doctorpod (approved):
My parting thought is with the feature specs which have become more complex - we should def think about how to make them more readable soon - but apart from that all good to go. Nice work!

PR #507, opened 2018-07-10
Add substance supply details

doctorpod:
Ticket: https://trello.com/c/VI1HpFyr

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_07_09_162410) do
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -234,6 +234,8 @@
     t.text "conceals_mobile_phone_or_other_items_details"
     t.string "violent_or_dangerous"
     t.text "violent_or_dangerous_details"
+    t.text "csra_details"
```
cesidio: csra_details should not be here

cesidio (changes requested):
fix schema.rb and all the rest is good to go

cesidio (approved):

PR #508, opened 2018-07-10
Override NTA font throughout with Arial... ugh

axemonkey:
Because the service is [not on GOV.UK](https://www.gov.uk/service-manual/design/making-your-service-look-like-govuk#if-your-service-isnt-on-govuk) we need to change the font to Arial throughout.

😢 

Before
---
image

After
---
image


axemonkey:
Before
---
image

After
---
image

cesidio (approved):

PR #519, opened 2018-07-24
Add sergeant approval process

cesidio:


app/controllers/escorts_controller.rb
```diff
@@ -30,6 +30,11 @@ def cancel
     end
   end
 
+  def approve
```
doctorpod, line 33: Consider a pre trigger at the top to authenticate sergeants maybe

doctorpod (approved):

app/controllers/escorts_controller.rb
cesidio: 👍

PR #531, opened 2018-08-09
Add menu highlighting on scroll to help screen

doctorpod:
Ticket: https://trello.com/c/HrC8TKcE


axemonkey (approved):
LGTM. 👍 :shipit: 

In a later update I'll change the Stickyfill polyfill init to use a generic classname, and then we can make this work in IE8 too.

PR #543, opened 2018-09-06
New design

cesidio:


app/models/user.rb
```diff
@@ -51,7 +51,7 @@ def authorized_establishments
   end
 
   def establishment(session)
-    return authorized_establishments.first if prison_officer?
+    return authorized_establishments.first
```
cesidio: This is wrong, you did commit something different from what was in master

PR #550, opened 2018-09-26
Add MIT Licence

matthewrudy:
According to MOJ Technical Guidance
each repo should have an MIT licence in the root directory
https://ministryofjustice.github.io/technical-guidance/standards/licencing-software-or-code/#each-repository-should-include-a-licence-file

(the guidance also says it should be in a file named `LICENCE` not `LICENSE` but it seems noone does that,
possibly because github doesn't recognise it?)


matthewrudy:
going to reopen from a branch on this repo

PR #551, opened 2018-09-26
compare_auto_alerts: store risk_cache.json in tmp

matthewrudy:
Not sure if this is the right solution.

But it seems strange to run the tests,
then for `risk_cache.json` to suddenly appear in the root dir.

Putting it in `tmp/` seems a simple solution.


doctorpod:
Yes totally agree. Good call. 
&lt;CaptainMainwaring&gt;I was wondering how long it would take for someone to notice that&lt;/CaptainMainwaring&gt;

doctorpod (approved):

PR #552, opened 2018-09-26
spec: assert the flow of paths for filling out a per

matthewrudy:
Saw this when just browsing through recent history.

Was interested to see whether we were `editing` an existing detainee, or `creating` a new one.

Seems the answer was `editing`.


doctorpod (approved):
Looks good - please rebase to master and merge

PR #555, opened 2018-09-26
Replace detainee ethnicity with type-ahead

doctorpod:
Use standard 16+1 ethnicity list.
Ticket: https://trello.com/c/Pbl1fCjP

### DIFF
--- a/app/models/detainee.rb
+++ b/app/models/detainee.rb
@@ -10,4 +10,24 @@ class Detainee < ApplicationRecord
   def female?
     gender == 'female'
   end
+
+  ETHNICITY_CODES = [
+    'A1 - Asian or Asian British Indian',
+    'A2 - Asian or Asian British Pakistani',
+    'A3 - Asian or Asian British Bangladeshi',
+    'A9 - Asian Other',
+    'B1 - Black or Black Caribbean',
+    'B2 - Black or Black African',
+    'B3 - Black Other',
+    'M1 - Mixed White & Black Caribbean',
+    'M2 - Mixed White & Black African',
+    'M3 - Mixed White and Asian',
+    'M9 - Mixed Other',
+    'NS - Not Stated',
+    'O1 - Chinese',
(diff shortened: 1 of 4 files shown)


cesidio (approved):

app/models/detainee.rb
```diff
@@ -10,4 +10,24 @@ def age
   def female?
     gender == 'female'
   end
+
+  ETHNICITY_CODES = [
+    'A1 - Asian or Asian British Indian',
```
matthewrudy, line 15: @doctorpod as a question,
would it be better to store the code (eg. `"A2"`) in the database

then map the code to the string in the view (eg. `"Asian or Asian British Pakistani"`)

The codes are not likely to change,
but I think some of those strings are quite odd.

doctorpod: Not really. If the codes or strings change it's just a case of editing the file. If they were stored in the DB we'd have to use a migration or datafix. I think it would be more involved.

matthewrudy: Maybe I'm misunderstanding,

but aren't you effectively storing in the database?

``` ruby
>> detainee.ethnicity = "A2 - Asian or British Pakistani"
>> detainee.save

>> puts detainee.ethnicity
A2 - Asian or British Pakistani
```

I'm suggesting it would have been better to store:

``` ruby
class Detainee # a presenter or whatever
  ETHNICITY_CODES = { "A2" =>  "A2 - Asian or British Pakistani"}
  def ethnicity_string
     ETHNICITY_CODES.fetch(ethnicity, ethnicity)
  end
end

>> detainee.ethnicity = "A2"
>> detainee.save

>> puts detainee.ethnicity
A2
>> puts detainee.ethnicity_string
A2 - Asian or British Pakistani
```

ie. the `code` is `A2`
the description of it `"A2 - Asian or British Pakistani"` is just part of the view layer, and subject to change.

PR #556, opened 2018-09-26
gems: bundle update

matthewrudy:
Went through and pinned a few versions

* rubocop - to avoid deciding on new rules
* wkhtmltopdf-binary - to avoid breaking the pdf tests

then just did an all round `bundle update`


matthewrudy:
if we're happy with this, should probably turn on dependabot
and just do daily dependency updates that way.

Gemfile.lock
```diff
@@ -103,13 +103,13 @@ GEM
...
       ffi (~> 1.0, >= 1.0.11)
-    chromedriver-helper (1.2.0)
+    chromedriver-helper (2.1.0)
       archive-zip (~> 0.10)
       nokogiri (~> 1.8)
     climate_control (0.2.0)
     cliver (0.3.2)
-    codeclimate-test-reporter (1.0.8)
-      simplecov (<= 0.13)
+    codeclimate-test-reporter (1.0.7)
```
matthewrudy: this rolled back because of the simplecov upgrade
but doesn't seem significant.

spec/support/fixtures/police/pdf-text-all-no-answers.txt
```diff
@@ -1,5 +1,5 @@
 Outbound vehicle
-Registration                     Cell number                        PECS Barcode
+Registration                     Cell number                       PECS Barcode
```
matthewrudy: I just eyeballed these changes.

Assuming its related to PDF parsing
not PDF generation.

Don't know how you normally handle that.

cesidio:
At the moment me and Clive are working to get rid of the deprecated front-end gems:

https://github.com/ministryofjustice/moving-people-safely/pull/543

matthewrudy:
@cesidio that's fine,
just pinned it to avoid it doing an update to 8.x

I rebased my changes to no longer touch any of the govuk gems.

matthewrudy:
@cesidio would be good to take another look

matthewrudy:
closing in favour of #572

PR #562, opened 2018-10-04
Add special vehicle details to move

doctorpod:
These are also cloned in re-use
Ticket: https://trello.com/c/5ESF0yZi

### DIFF
--- a/app/controllers/escorts_controller.rb
+++ b/app/controllers/escorts_controller.rb
@@ -12,8 +12,7 @@ class EscortsController < ApplicationController
   end
 
   def create
-    escort = EscortCreator.call(escort_params)
-    escort.create_move(from_establishment: current_user.establishment(session))
+    escort = EscortCreator.call(escort_params, current_user.establishment(session))
     EscortPopulator.new(escort).call if escort.from_prison?
     if escort.detainee
       redirect_to edit_escort_detainee_path(escort)
--- a/app/services/escort_creator.rb
+++ b/app/services/escort_creator.rb
@@ -12,8 +12,8 @@ class EscortCreator
   def call
     if existent_escort
       deep_clone_escort.tap do |clone|
-        build_move_with_special_vehicle_details(clone)
         clone.twig = existent_escort
+        clone.move.from_establishment = from_establishment
         clone.needs_review!
       end
     else
@@ -29,6 +29,7 @@ class EscortCreator
 
   INCLUDE_GRAPH = [
     :detainee,
+    :move,
     { risk: [:must_not_return_details] },
     { healthcare: [:medications] },
     :offences,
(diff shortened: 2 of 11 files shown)


app/services/escort_creator.rb
```diff
@@ -54,4 +58,14 @@ def deep_clone_escort
       except: EXCEPT_GRAPH
     )
   end
+
+  def build_move_with_special_vehicle_details(escort)
```
cesidio: This is not consistent with the approach we use when cloning an existing escort. I would add the move to INCLUDE_GRAPH, and all the fields we are not interested in EXCEPT_GRAPH.

cesidio (changes requested):

app/controllers/escorts_controller.rb
```diff
@@ -12,8 +12,7 @@ def show
   end
 
   def create
-    escort = EscortCreator.call(escort_params)
-    escort.create_move(from_establishment: current_user.establishment(session))
+    escort = EscortCreator.call(escort_params, current_user.establishment(session))
```
cesidio, line 15: 👍 makes sense to create the move in the EscortCreator

app/services/escort_creator.rb
doctorpod: Willdo

spec/services/escort_creator_spec.rb
```diff
@@ -146,6 +164,15 @@ def expect_offences_to_be_cloned(existent_escort, new_escort)
     expect(offences_attributes).to match_array(expected_offences_attributes)
   end
 
+  def expect_move_special_vehicle_details_to_be_cloned(existent_escort, new_escort)
+    %i[require_special_vehicle require_special_vehicle_details
```
cesidio: To be consistent with the rest of the spec, I would add this attributes to the method:
```
except_move_attributes
```

spec/services/escort_creator_spec.rb
```diff
@@ -146,6 +164,15 @@ def expect_offences_to_be_cloned(existent_escort, new_escort)
     expect(offences_attributes).to match_array(expected_offences_attributes)
   end
 
+  def expect_move_special_vehicle_details_to_be_cloned(existent_escort, new_escort)
```
cesidio: Also for consistency, I would make this method similar to what we do for the cloned detainee:

```
  def expect_move_to_be_cloned(existent_escort, new_escort)
    move_attributes = existent_escort.move.attributes.except(*except_move_attributes)
    expect(new_escort.move.id).not_to eq(existent_escort.move.id)
    expect(new_escort.move).to have_attributes(move_attributes)
  end
```

cesidio (approved):

PR #569, opened 2018-10-24
Update police suicide and add observation level question

doctorpod:
Ticket: https://trello.com/c/GTN5DnB0


db/migrate/20181031141018_split_observation_level_details.rb
```diff
@@ -0,0 +1,8 @@
+class SplitObservationLevelDetails < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :risks, :observation_level_details, :text
+    add_column :risks, :observation_level2_details, :text
```
cesidio: We need to design a better solution, no point in having 3 fields in the database of which only one is going to be filled in. We can fix this not replicating 3 textfields in the view with a simple javascript functionality that hide/show the only required one. Or in the worst case we can add some logic in the form to populate and validate the only field that is going to be persisted in the database.

cesidio (changes requested):

db/migrate/20181031141018_split_observation_level_details.rb
doctorpod: I think the best way forward to revert back to a single observation_level_detail attribute is the form approach as I already spent a lot of time trying to wrestle the new Gov UK JS into the shape we want without success.

cesidio (approved):

PR #570, opened 2018-10-25
Detainee mandatory fields and minor changes

cesidio:


app/views/detainees/_form.html.slim
```diff
@@ -33,7 +33,7 @@
           legend_options: { class: 'govuk-label' }
 
       .govuk-form-group
-        label.govuk-label for='auto_ethnicity' Ethnicity
+        label.govuk-label for='auto_ethnicity' Ethnicity #{' (Optional)' if escort.from_police?}
```
doctorpod, line 36: Probably a better I18n way to do this; `location_text` and `localized_label` come to mind - can't quite remember how they work but I think they address this sort of scenario.

PR #577, opened 2018-11-07
Use of partials in application layout

cesidio:


doctorpod (approved):
Very nice separation of concerns

PR #578, opened 2018-11-07
Bundle update

cesidio:


matthewrudy:
I think if its a plain bundle-update
which doesn't break tests.
Should just merge these straight away.

PR #595, opened 2018-11-14
IE8 warning banner

doctorpod:
Ticket: https://trello.com/c/7Z9GDg3i

### DIFF
--- a/app/views/layouts/_main.html.slim
+++ b/app/views/layouts/_main.html.slim
@@ -13,4 +13,5 @@
         div class="alert alert--#{severity}" = flash_message_for(flash[severity])
 
   main#main-content.govuk-main-wrapper role="main"
+    = render 'shared/browser_warning' if request.env['HTTP_USER_AGENT'] =~ /MSIE 8/
     = yield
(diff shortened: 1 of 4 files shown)


cesidio (approved):

app/views/layouts/_main.html.slim
```diff
@@ -13,4 +13,5 @@
         div class="alert alert--#{severity}" = flash_message_for(flash[severity])
 
   main#main-content.govuk-main-wrapper role="main"
+    = render 'shared/browser_warning' if request.env['HTTP_USER_AGENT'] =~ /MSIE 8/
```
matthewrudy, line 16: I looked through the list of common user agents,
and it seems this is good enough.
(unless they run IE8 in IE6 or IE7 compatibility mode??)

https://developers.whatismybrowser.com/useragents/explore/software_type_specific/web-browser/6

matthewrudy:
Extra question:

It says we won't support "this version of Internet Explorer", ie. IE 8
which I believe is correct.

But then it tells you to use Chrome or Firefox.
But Quantum is supposed to be shipping IE 11.
Do we want to mention that?

use-a-different-browser

PR #596, opened 2018-11-14
Add filename to PDF download

matthewrudy:
Ticket: https://trello.com/c/o5On82Gx/520-at-medway-and-at-salfords-custodies-the-pdf-does-not-appear-in-a-separate-tab

Keep it inline,
but let the browser know what filename to use.

For a modern browser this means:

* when it's open in a tab, if you click to save, it uses the filename
* you can right click "save as", and it will use the filename

Hopefully this will fix the bug seen in Police.

BONUS:

fixed a couple of bugs I found on the way:

* avoid an error when trying to create a per with no offences
* fix some missing translations
* ensure uploaded pdfs have the .pdf extension

### DIFF
--- a/app/services/escort_issuer.rb
+++ b/app/services/escort_issuer.rb
@@ -38,7 +38,7 @@ class EscortIssuer
 
   def generate_per_document
     pdf = PdfGenerator.new(escort).call
-    file = Tempfile.new('per')
+    file = Tempfile.new(['per', '.pdf'])
     file.write(pdf.force_encoding('utf-8'))
     file
   end
--- a/app/services/escorts/generate_documents_for_issued_pers.rb
+++ b/app/services/escorts/generate_documents_for_issued_pers.rb
@@ -40,7 +40,7 @@ module Escorts
 
       def generate_per_document
         pdf = PdfGenerator.new(escort).call
-        file = Tempfile.new('per')
+        file = Tempfile.new(['per', '.pdf'])
         file.write(pdf.force_encoding('utf-8'))
         file
       end
--- a/spec/features/printing_prison_per_spec.rb
+++ b/spec/features/printing_prison_per_spec.rb
@@ -85,6 +85,12 @@ RSpec.feature 'printing a prison PER', type: :feature do
       visit escort_path(escort)
       escort_page.click_print
 
+      escort.reload
+      expected_filename = escort.document_file_name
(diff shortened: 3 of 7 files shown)


spec/features/printing_prison_per_spec.rb
```diff
@@ -85,6 +85,12 @@
       visit escort_path(escort)
       escort_page.click_print
 
+      escort.reload
+      expected_filename = escort.document_file_name
+      expect(expected_filename).to match(/^per[0-9a-z-]+\.pdf$/i)
+      expect(page.response_headers['Content-Type']).to eq 'application/pdf'
+      expect(page.response_headers['Content-Disposition']).to eq "inline; filename=#{expected_filename.inspect}"
```
matthewrudy, line 92: I did think to throw this in a matcher,
but decided that felt a bit heavy.

app/services/escorts/generate_documents_for_issued_pers.rb
```diff
@@ -40,7 +40,7 @@ def issued_per_document
 
       def generate_per_document
         pdf = PdfGenerator.new(escort).call
-        file = Tempfile.new('per')
+        file = Tempfile.new(['per', '.pdf'])
```
matthewrudy, line 43: there are no tests on this that I can see.

app/services/escort_issuer.rb
```diff
@@ -38,7 +38,7 @@ def issued_per_document
 
   def generate_per_document
     pdf = PdfGenerator.new(escort).call
-    file = Tempfile.new('per')
+    file = Tempfile.new(['per', '.pdf'])
```
matthewrudy, line 41: ideally this would actually use a filename we wanted
rather than a random one.

But decided that was out of scope of this quick fix.

matthewrudy:
Can't verify how this affects ie8
In my vm both this and master try to download the pdf, rather than rendering inline.

cesidio:
I like these changes. The only thing I would suggest is to test this branch on staging, and check that everything works as expected: new PERs (this should work) and old PERs (the link to the document on AWS should not be affected, but let's test it)

cesidio (approved):

matthewrudy:
@cesidio sure
will test before and after merge on this example:
https://mps-staging.dsd.io/escorts/38e270c3-12ea-486c-9281-2e2a69d31bab/print/new

PR #598, opened 2018-11-14
gems: always use github: 'user/repo' in the gemfile

matthewrudy:
more succinct than specifying the full git path each time


Gemfile.lock
```diff
@@ -1,5 +1,5 @@
 GIT
-  remote: https://github.com/thethanghn/custom-err-msg.git
+  remote: https://github.com/thethanghn/custom-err-msg
```
matthewrudy, line 2: it annoying that this changed
but doesn't make any difference

cesidio (approved):

PR #617, opened 2018-11-22
Police healthcare inconsistencies

doctorpod:
Ticket: https://trello.com/c/ayssIGv4

### DIFF
--- a/app/controllers/healthcare_controller.rb
+++ b/app/controllers/healthcare_controller.rb
@@ -1,9 +1,5 @@
 class HealthcareController < AssessmentsController
-  before_action :skip_intro_if_police, only: %i[intro]
-
   private
-
-  include WizardHelper
 
   def model
     Healthcare
@@ -19,9 +15,5 @@ class HealthcareController < AssessmentsController
 
   def show_page
     escort_healthcare_path(escort)
-  end
-
-  def skip_intro_if_police
-    redirect_to assessment_wizard_path(assessment) if assessment.location == 'police'
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/healthcare_controller.rb
```diff
@@ -1,6 +1,10 @@
 class HealthcareController < AssessmentsController
+  before_action :skip_intro_if_police, only: %i[intro]
```
cesidio: Let's not introduce a before_action for that. This can be done by just changing the not_started_path value of the options passed to render the partial in ```app/views/escorts/show.html.slim```
```
    = render 'assessment', assessment: escort.risk,
                           model: Risk,
                           not_started_path: (escort.from_prison? ? new_escort_risks_path(escort) : intro_escort_risks_path(escort)),
                           summary_path: escort_risks_path(escort),
                           section: escort.risk || escort.build_risk
```

Here you can see that for prisons we do not show the intro page for the risk assessment.

cesidio (changes requested):

cesidio (approved):

app/form_builders/mps_form_builder.rb
```diff
@@ -201,16 +201,23 @@ def field_classes(attribute, method_name = :text_area)
...
-    key = "#{name}.#{attribute}"
+    key = "#{flattened_object_name}.#{attribute}"
     location_text(translate(key, scope))
   end
 
+  def flattened_object_name
+    object_name.gsub(/\[(.*)_attributes\]\[\d+\]/, '.\1')
+  end
+
+  def default_label(attribute)
```
cesidio: default_label method has been removed

cesidio (changes requested):

PR #618, opened 2018-11-22
Cleanup form builder

cesidio:


doctorpod (approved):
Nice work. Good to see further standardisation in line with the gov helpers.

PR #625, opened 2018-11-27
Custody API Playground

matthewrudy:
[Trello Card](https://trello.com/c/FNg5iCnW/516-create-a-harness-to-allow-us-to-use-nomis-api-to-view-prison-moves)

If we set the credentials on staging,
Paul can play with this.

Looks like this:

Fill in the NOMS ID:
screenshot 2018-11-27 17 17 47

If there's a result:
screenshot 2018-11-27 17 18 00

If there isn't:
screenshot 2018-11-27 17 18 09


matthewrudy:
@cesidio this all works
but don't know what changes you would like to have it added to the codebase.

In particular:

* the controller is very scripty
* that `bin` feels out of place
* maybe remove the test values into some other file
* add a VCR test or something?

PR #642, opened 2018-11-30
Add slim_lint gem

cesidio:


.slim-lint.yml
```diff
@@ -0,0 +1,3 @@
+linters:
+  LineLength:
+    max: 500
```
matthewrudy: I did some digging into why you set this,
and see its because of the inline svg in the footer.

If you set it to 150 this is the output.
```
app/views/escorts/print/cover.html.slim:136 [W] LineLength: Line is too long. [256/150]
app/views/homepage/help.html.slim:240 [W] LineLength: Line is too long. [154/150]
app/views/layouts/_footer.html.slim:10 [W] LineLength: Line is too long. [161/150]
app/views/layouts/_footer.html.slim:11 [W] LineLength: Line is too long. [477/150]
app/views/layouts/_footer.html.slim:18 [W] LineLength: Line is too long. [210/150]
```

I think we can try:
```
linters:
  LineLength:
    max: 150
    exclude:
      - app/views/layouts/_footer.html.slim
```
then just fix the 2 remaining issues.

cesidio: Just added your suggestions, thanks.

PR #643, opened 2018-12-01
Add scss-lint gem

cesidio:


matthewrudy:
I think this is fine.
Some of their defaults I disagree with.

eg.

Double Quotes:
In your ruby you favour `'red'` over `"red"`
but in sass, it favours `"red"` over `'red'`

Leading Zeros:
How is `.5em` better than `0.5em`

https://github.com/brigade/scss-lint/blob/ee93a0ede61992d272d5c186a9305e38ab0f2499/lib/scss_lint/linter/README.md#leadingzero

Otherwise, don't see any problems with it.

matthewrudy (approved):

PR #645, opened 2018-12-03
Fix style and errors in forms with nested objects

cesidio:


matthewrudy (approved):
@cesidio it's hard for me to know the subtlety and consequences of these changes.

Particularly the changes to `multiples.js`.

I notice that there's still a reference to `.multiple-choice` in the `main-ie8.scss` which is probably worth deleting.

Apart from this,
I can just approve it as looking generally fine.

cesidio:
Thank you for the review. I removed the reference to .multiple-choice.

PR #646, opened 2018-12-03
Improve rubocop

cesidio:


matthewrudy (approved):
I'm a big fan of frozen string literals.

PR #698, opened 2019-01-08
Require details on all observation levels

doctorpod:
Ticket: https://trello.com/c/zN3vJVuW

### DIFF
--- a/app/helpers/risks_helper.rb
+++ b/app/helpers/risks_helper.rb
@@ -1,14 +1,6 @@
 # frozen_string_literal: true
 
 module RisksHelper
-  def heightened_observation_yes_no(level)
-    if Forms::Risk::RiskToSelf::OBSERVATION_LEVELS_REQUIRING_DETAILS.include?(level)
-      'yes'
-    else
-      'no'
-    end
-  end
-
   def full_observation_level_details(level, details)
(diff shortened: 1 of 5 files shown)


app/helpers/risks_helper.rb
```diff
@@ -1,14 +1,6 @@
...
 module RisksHelper
-  def heightened_observation_yes_no(level)
-    if Forms::Risk::RiskToSelf::OBSERVATION_LEVELS_REQUIRING_DETAILS.include?(level)
-      'yes'
-    else
-      'no'
-    end
-  end
-
   def full_observation_level_details(level, details)
```
cesidio: What is the point of full_observation_level_details method now? Can we remove this method as well?

cesidio (approved):

app/helpers/risks_helper.rb
```diff
@@ -1,14 +1,6 @@
...
     return '' if level == Forms::Risk::RiskToSelf::OBSERVATION_LEVELS.first
```
cesidio: This line:

```
return '' if level == Forms::Risk::RiskToSelf::OBSERVATION_LEVELS.first
```

in method full_observation_level_details should be now removed as a result of your work

cesidio (changes requested):

PR #711, opened 2019-01-10
Fix detainee image error on court receiver view

cesidio:


doctorpod (approved):
As long as it works looks fine to me.

PR #795, opened 2019-03-15
Show dash when Not for release is no in police

cesidio:


app/helpers/escorts_helper.rb
```diff
@@ -62,7 +62,8 @@ def from(move)
   end
 
   def not_for_release_text(move)
-    return 'Contact the prison to confirm release' if move.not_for_release == 'no'
+    return 'Contact the prison to confirm release' if move.not_for_release == 'no' && move.from_prison?
+    return '-' if move.not_for_release == 'no' && move.from_police?
```
doctorpod: There's a dash helper method in ApplicationHelper - it might be nice to use this but not a biggie.

doctorpod (approved):

PR #1196, opened 2020-05-20
Fix variable naming in handling of non-json Nomis response

willfish:
### What?

Nomis is returning something that isn't JSON when trying to populate detainee information.

Unfortunately we've made a typo that means we can't see the response that is getting returned by the client. This PR fixes that typo and adds some specs to avoid regressions.

### Why?

So we can get more information back from the Nomis client.

### A/C

Existing tests still pass.
Jason is happy.


jason-sims (approved):
👍

